PR #23, opened 2020-10-13
Remove Redis storage adapter

bonustrack:


.env
```diff
@@ -1,7 +1,6 @@
 PORT=8080
 NETWORK=testnet
-DATABASE_URL=redis://...
-JAWSDB_URL=mysql://...
+JAWSDB_URL=mysql://<user>:<pwd>@<host>:3306/
```
fforbeck, line 3:
```suggestion
JAWSDB_URL=mysql://<user>:<pwd>@<host>:3306/<db-name>
```

fforbeck:
@bonustrack 
I noticed you have used the `KEY` keyword to create the indexes for the MySQL tables. Was wondering if we can change all these `KEY` declarations (https://github.com/balancer-labs/snapshot-hub/blob/208ba37c80419bb866db2dff3db1655b14079606/server/helpers/database/schema.sql#L13) to the `INDEX` keyword, that will make it compatible with Postgres as well

PR #24, opened 2020-10-13
Init MySQL db with default schema.sql

fforbeck:
It adds the init script for MySQL, so the schema and tables are created when the DB container is started.
 
- Added sample `JAWSDB_URL` for mysql
- Added `MYSQL_ROOT_PASSWORD` to fix MySQL start up error 
- Added new volume to docker-compose to init DB with default schema

Screenshot from 2020-10-13 14-39-33


bonustrack (approved):
Looks great! Thank you for your PR!

PR #25, opened 2020-10-13
Make the schema compatible with Postgres

fforbeck:
- Using `INDEX` keyword to create the column indexes.
- Removed back-tic ` `` ` to solve SQL error in Postgres (the column names are always created in lower case).

The same schema.sql can be used for MySQL and Postgres - tested both.


bonustrack:
Looks good!

PR #35, opened 2021-01-28
feat: Delete proposal

samuveth:
Fixes https://github.com/balancer-labs/snapshot/issues/214

- When message is `delete-proposal` change the proposal type to `archive-proposal`.
- We might need to increase the `type VARCHAR(12) NOT NULL,` to fit it.
-

### DIFF
--- a/server/helpers/adapters/mysql.ts
+++ b/server/helpers/adapters/mysql.ts
@@ -4,40 +4,54 @@ import fleek from '@fleekhq/fleek-storage-js';
 import { isAddress, getAddress } from '@ethersproject/address';
 import db from '../mysql';
 
-export async function storeProposal(space, body, authorIpfsHash, relayerIpfsHash) {
+export async function storeProposal(
+  space,
+  body,
+  authorIpfsHash,
+  relayerIpfsHash
+) {
   const msg = JSON.parse(body.msg);
   let query = 'INSERT IGNORE INTO messages SET ?;';
-  await db.queryAsync(query, [{
-    id: authorIpfsHash,
-    address: body.address,
-    version: msg.version,
-    timestamp: msg.timestamp,
-    space,
-    type: 'proposal',
-    payload: JSON.stringify(msg.payload),
-    sig: body.sig,
-    metadata: JSON.stringify({
-      relayer_ipfs_hash: relayerIpfsHash
-    })
-  }]);
+  await db.queryAsync(query, [
+    {
+      id: authorIpfsHash,
+      address: body.address,
+      version: msg.version,
+      timestamp: msg.timestamp,
+      space,
+      type: 'proposal',
+      payload: JSON.stringify(msg.payload),
+      sig: body.sig,
+      metadata: JSON.stringify({
+        relayer_ipfs_hash: relayerIpfsHash
+      })
+    }
+  ]);
+}
+
+export async function archiveProposal(authorIpfsHash) {
+  let query = 'UPDATE messages SET type = ? WHERE id = ?';
+  await db.queryAsync(query, ['archive-proposal', authorIpfsHash]);
 }
 
 export async function storeVote(space, body, authorIpfsHash, relayerIpfsHash) {
   const msg = JSON.parse(body.msg);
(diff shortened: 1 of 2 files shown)


server/helpers/adapters/mysql.ts
```diff
@@ -4,40 +4,54 @@ import fleek from '@fleekhq/fleek-storage-js';
...
+      sig: body.sig,
+      metadata: JSON.stringify({
+        relayer_ipfs_hash: relayerIpfsHash
+      })
+    }
+  ]);
+}
+
+export async function archiveProposal(authorIpfsHash) {
+  let query = 'UPDATE messages SET type = ? WHERE id = ?';
```
bonustrack, line 33: It's better to add a LIMIT here
```suggestion
  let query = 'UPDATE messages SET type = ? WHERE id = ? LIMIT 1';
```
Also it's better to do a WHERE type = 'proposal'

server/index.ts
```diff
@@ -106,50 +114,78 @@ router.post('/message', async (req, res) => {
...
+  if (
+    !(await verifySignature(
+      body.address,
+      body.sig,
+      hashPersonalMessage(body.msg)
+    ))
+  )
     return sendError(res, 'wrong signature');
 
+  if (msg.type === 'delete-proposal') {
```
bonustrack, line 148: Can you move this verification to here https://github.com/snapshot-labs/snapshot-hub/pull/35/files#diff-786114c6ebf98e577ec30dac884800bb7c9ce800b5323710ed03b916e1d33b68R240 ?

server/index.ts
```diff
@@ -106,50 +114,78 @@ router.post('/message', async (req, res) => {
...
+    let query = 'SELECT address FROM messages WHERE id = ?';
```
bonustrack, line 149: You should explicitly check for type = 'proposal', otherwise it could be used to delete vote

samuveth: So the query can be changes by the user?

bonustrack: Someone could build a signed message to delete a vote of himself. We dont want this, when a proposal ended the results should not change.

samuveth: Like this e0423ba?

bonustrack: Yes

server/index.ts
```diff
@@ -194,6 +227,17 @@ router.post('/message', async (req, res) => {
     version: '2'
   });
 
+  if (msg.type === 'delete-proposal') {
+    let query = `SELECT address FROM messages WHERE type = 'proposal' AND id = ?`;
+    let propasalSigner = await db.queryAsync(query, [msg.payload.proposal]);
+    if (
+      propasalSigner[0].address.toLowerCase() !== body.address.toLowerCase()
```
bonustrack, line 234: The address is never lowercase, you could update the query to use `WHERE address = ?` to tackle this

samuveth: I couldn't get that to work with `WHERE address = ?` it gives me undefined error that I can't manage to fix. I've removed the `toLowerCase()` now, I'm not sure anymore why I needed them before.

bonustrack: Hum that's weird

server/index.ts
```diff
@@ -194,6 +227,15 @@ router.post('/message', async (req, res) => {
     version: '2'
   });
 
+  if (msg.type === 'delete-proposal') {
+    let query = `SELECT address FROM messages WHERE type = 'proposal' AND id = ?`;
+    let propasalSigner = await db.queryAsync(query, [msg.payload.proposal]);
+    if (propasalSigner[0].address !== body.address) {
+      return sendError(res, 'wrong signer');
+    }
+    await archiveProposal(msg.payload.proposal);
```
bonustrack, line 236: Actually i mistaken in my previous comment, you was right, the `SELECT` query to check if the action is valid should happen before the `pinJson` function and the `archiveProposal` should happen after, so we need `if (msg.type === 'delete-proposal') {` in 2 different place. Sorry for the hassle!

bonustrack (approved):
Looking good

PR #41, opened 2021-03-07
Vue cli to typescript

### LINKED ISSUE #39: Remove vue cli pkg and use tsc instead
We should remove `vue-cli-service` and all the pkg related to Vue and run TS with `tsc` directly
https://github.com/snapshot-labs/snapshot-hub/blob/master/package.json#L7-L10

ChaituVR:
Fixes #39 

also make sure to merge https://github.com/snapshot-labs/snapshot-spaces/pull/942, before testing this


docker-compose.yml
```diff
@@ -6,7 +6,7 @@ services:
     build: .
     container_name: snapshot-hub
     env_file:
-      - .env.local
+      - .env
```
bonustrack, line 9: Any reason you change the file here?

server/helpers/mysql.ts
```diff
@@ -4,6 +4,8 @@ import Connection from 'mysql/lib/Connection';
 import bluebird from 'bluebird';
 import parse from 'connection-string';
 
+require('dotenv').config()
```
bonustrack, line 7: I think you don't need to import this twice, the config vars should be already available from /index.ts isn't?

bonustrack:
@ChaituVR Also deployment is failing on GitHub action you may want to look on this.

ChaituVR:
@bonustrack once https://github.com/snapshot-labs/snapshot-spaces/pull/1101 this is merged, will change package.json to take `master` branch

bonustrack:
@ChaituVR Ok thanks it's merged

index.ts
```diff
@@ -1,13 +1,20 @@
+require('dotenv').config()
+
 import bodyParser from 'body-parser';
 import frameguard from 'frameguard';
 import cors from 'cors';
 import api from './server';
+import express from 'express';
+
+
```
bonustrack, line 9:
```suggestion
```

server/helpers/relayer.ts
```diff
@@ -1,5 +1,11 @@
 import { Wallet } from '@ethersproject/wallet';
 
+declare var process : {
```
bonustrack, line 3: Any reason we need this here but not on others place where we use `process.env...`? Can't this be handled with the config files so we don't need to add types everytime?

ChaituVR: @bonustrack removed it.. it is interpreting the value as undefined ..

package.json
```diff
@@ -4,10 +4,12 @@
   "repository": "balancer-labs/snapshot-hub",
   "license": "MIT",
   "scripts": {
-    "lint": "vue-cli-service lint ./server/**/*.ts",
-    "express": "vue-cli-service express:watch",
-    "express:run": "vue-cli-service express:run",
-    "start": "vue-cli-service express:run"
+    "lint": "eslint ./server/**/*.ts --fix",
```
bonustrack, line 7: This script dont work for me, it does for you?

.gitignore
```diff
@@ -1,6 +1,7 @@
 .DS_Store
 node_modules
 /dist
+/**/dist/**.js
```
bonustrack, line 4: Just wonder how is this being used, i don't see any file generated on a dist folder.

package.json
ChaituVR: ya working for me.. you are getting some error?

bonustrack: I mean, i introduce some lint error and run the script but nothing got fixed. This doesn't happen for you?

ChaituVR: @bonustrack fixed it .. please review now

bonustrack (approved):
Looks good!

PR #49, opened 2021-04-15
Update proposal.ts

Herschee:
- honor new configuration field `filters.exclusions` for addresses to be excluded from a space's proposals  

e.g 
`{
  "token": ..
  "strategies": [
    {
     ..
      }
    }
  ],
  "members": [],
  "filters": {
    "exclusions": []
    }
}`

### DIFF
--- a/server/writer/proposal.ts
+++ b/server/writer/proposal.ts
@@ -43,8 +43,8 @@ export async function verify(body): Promise<any> {
   const members = space.members
     ? space.members.map(address => address.toLowerCase())
     : [];
-  const exclusions = space.filter.exclusions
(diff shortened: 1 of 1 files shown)


bonustrack:
How would you like to use this feature? I feel like having blacklisting will be complicated to manage because anyone can generate a new address, it will be easy to bypass this limitation.

Herschee:
> How would you like to use this feature? I feel like having blacklisting will be complicated to manage because anyone can generate a new address, it will be easy to bypass this limitation.

sorry for the delay - this is no longer our intended directive. it's been updated more towards a `minBalance` requirement for creating proposals. i will abandon this PR

PR #54, opened 2021-04-21
Call getProfiles only when profile data is requested

ChaituVR:


server/graphql/schema.gql
```diff
@@ -11,11 +11,16 @@ type Space {
...
+  profile: Profile
+}
+
+type Profile {
   ens: String
   name: String
   image: String
 }
 
+
```
bonustrack, line 23:
```suggestion
```

PR #57, opened 2021-04-23
Vote Query

ChaituVR:
image

### DIFF
--- a/server/graphql/index.ts
+++ b/server/graphql/index.ts
@@ -85,6 +85,46 @@ export const rootValue = {
           space
         };
       });
+    },
+    votes: async (
+      parent,
+      { 
+        id,
+        first = 10,
+        skip = 0,
+        address,
+        space
+      },
+      context,
+      info
+    ) => {
+      let queryStr = "";
+      const params: any[] = [];
+      if (id) {
+        queryStr += `AND id = ? `;
+        params.push(id);
+      }
+      if (address) {
+        queryStr += `AND address = ? `;
+        params.push(address);
+      }
+      if (space) {
+        queryStr += `AND space = ? `;
+        params.push(space);
+      }
+      params.push(skip, first);
+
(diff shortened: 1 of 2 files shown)


server/graphql/schema.gql
```diff
@@ -33,6 +33,15 @@ type Proposal {
   snapshot: String
 }
 
+type Vote {
+  id: String
+  timestamp: Int
+  spaceId: String
```
bonustrack, line 39: We should use a Space entity here

server/graphql/schema.gql
```diff
@@ -33,6 +33,15 @@ type Proposal {
...
+  proposalId: String
+  choice: Int
+  voteAddress: String
```
bonustrack, line 42: This should be `voter`. And voter should give you profile same as "author" in timeline like this ```
{
  voter {
    address
    profile {
      ens
      name
    }
  } 
}```

server/graphql/index.ts
```diff
@@ -85,6 +85,46 @@ export const rootValue = {
...
       });
+    },
+    votes: async (
+      parent,
+      { 
+        id,
+        first = 10,
+        skip = 0,
+        address,
+        space
```
bonustrack, line 96: This should be "spaces" array (same as timeline query). Also can you add an argument "proposal"? This way we can use this endpoint on the proposal page to replace the Rest API.

server/graphql/index.ts
```diff
@@ -85,6 +85,59 @@ export const rootValue = {
...
+      if (address) {
+        queryStr += `AND address = ? `;
+        params.push(address);
+      }
+      if (spaces.length) {
+        queryStr += `AND space IN (?) `;
+        params.push(spaces);
+      }
+      if (proposal) {
+        queryStr += `AND payload like '%"proposal": "${proposal}"%' `;
```
bonustrack, line 111: You should always `params` to pass external data so it get encoded otherwise someone could inject SQL queries..

bonustrack:
Closing this PR, we'll do another way (using "votes" table not view)

PR #60, opened 2021-04-30
Add rate limit

bonustrack:


server/index.ts
```diff
@@ -29,7 +30,7 @@ router.get('/', (req, res) => {
   });
 });
 
-router.get('/spaces/poke', async (req, res) => {
+router.get('/spaces/poke', rateLimit, async (req, res) => {
```
ChaituVR, line 33: this `rateLimit` middleware is not for all routes?

bonustrack: No, some route don't need it.

server/helpers/ratelimit.ts
```diff
@@ -0,0 +1,31 @@
...
+  if (!requests[key]) requests[key] = {};
+  if (!requests[key][id]) requests[key][id] = [];
+  requests[key][id] = requests[key][id].filter(
+    timestamp => timestamp > ts - interval
+  );
+  if (requests[key][id].length >= max) {
+    console.log('Rate limited', id);
+    return true;
+  }
+  requests[key][id].push(ts);
```
sohkai, line 19: It'd be better to handle this with a counter if possible, since otherwise this is just going to eat up memory.

`express-rate-limit` handles this pretty cheekily with [a global interval that resets all the keys](https://github.com/nfriedly/express-rate-limit/blob/master/lib/memory-store.js#L41-L44).

sohkai: Actually for posterity, I read l12 wrong 😅. The memory does continue to grow, but it's not as bad as I thought since it's bounded to a limited size array to each ip (I originally thought the array just continued to grow in size).

This approach is technically better at smoothing out requests over time, but the simple counter version should be enough for our needs.

PR #61, opened 2021-04-30
Rate limit with express-rate-limit

ChaituVR:


index.ts
```diff
@@ -2,15 +2,24 @@ import bodyParser from 'body-parser';
...
 export default app => {
   app.use(bodyParser.json({ limit: '20mb' }));
   app.use(bodyParser.urlencoded({ limit: '20mb', extended: false }));
   app.use(frameguard({ action: 'deny' }));
   app.use(cors());
+  app.use(rateLimit({
+    windowMs: 30 * 1000,
+    max: 120,
+    handler: (req, res) => {
+      sendError(res, 'rate limited')
```
bonustrack, line 20: Can you add a console.log('Rate limited'); ?

index.ts
```diff
@@ -2,15 +2,24 @@ import bodyParser from 'body-parser';
 import frameguard from 'frameguard';
 import cors from 'cors';
 import { graphqlHTTP } from 'express-graphql';
+import rateLimit from "express-rate-limit";
```
bonustrack, line 5:
```suggestion
import rateLimit from 'express-rate-limit';
```

index.ts
```diff
@@ -2,15 +2,25 @@ import bodyParser from 'body-parser';
...
 import { schema, rootValue } from './server/graphql';
 import defaultQuery from './server/graphql/examples';
+import { sendError } from './server/helpers/utils';
 
 export default app => {
   app.use(bodyParser.json({ limit: '20mb' }));
   app.use(bodyParser.urlencoded({ limit: '20mb', extended: false }));
   app.use(frameguard({ action: 'deny' }));
   app.use(cors());
+  app.use(rateLimit({
```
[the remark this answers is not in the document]
ChaituVR: Ahh yes.. sorry didn't notice this in the documentation .. added 👍

bonustrack:
@ChaituVR That should help based on the logs there is quite lot of requests at the same time, and i think it's better to start high so we can see the limitation of the hub, we can adjust if needed.

index.ts
```diff
@@ -2,15 +2,26 @@ import bodyParser from 'body-parser';
...
 import { schema, rootValue } from './server/graphql';
 import defaultQuery from './server/graphql/examples';
+import { sendError } from './server/helpers/utils';
 
 export default app => {
   app.use(bodyParser.json({ limit: '20mb' }));
   app.use(bodyParser.urlencoded({ limit: '20mb', extended: false }));
   app.use(frameguard({ action: 'deny' }));
   app.use(cors());
+  app.set('trust proxy', 1);
```
sohkai, line 16: It might be worth putting this behind a env var, since it wouldn't apply to local deployments.

index.ts
```diff
@@ -2,15 +2,26 @@ import bodyParser from 'body-parser';
...
+  app.use(rateLimit({
+    windowMs: 30 * 1000,
+    max: 120,
```
bonustrack, line 19:
```suggestion
    max: 12,
```

index.ts
```diff
@@ -2,15 +2,26 @@ import bodyParser from 'body-parser';
...
 
 export default app => {
   app.use(bodyParser.json({ limit: '20mb' }));
   app.use(bodyParser.urlencoded({ limit: '20mb', extended: false }));
   app.use(frameguard({ action: 'deny' }));
   app.use(cors());
+  app.set('trust proxy', 1);
+  app.use(rateLimit({
+    windowMs: 30 * 1000,
+    max: 12,
```
sohkai, line 19: I think 12 might be too low for 30s (you might open 10 tabs, for example). What do you think about making the window a bit shorter, e.g. 10s?

I think what we're seeing are very large, sudden bursts of requests from programmatic sources, so the shorter the window the less normal users could be impacted by it.

sohkai (approved):
🙌 🙌 🙌

index.ts
```diff
@@ -2,15 +2,26 @@ import bodyParser from 'body-parser';
...
+    handler: (req, res) => {
+      console.log('Rate limited');
+      sendError(res, 'rate limited')
+    }
```
sohkai, line 23: Ah, we may also want to use the [`keyGenerator`](https://www.npmjs.com/package/express-rate-limit#keygenerator) option to hash the ip.

Although @bonustrack, was there a reason for preferring this to just storing the raw ips in #60?

bonustrack: On #60 the idea was to get a hash of the IP so we can recognize same users in logs without exposing any IPs. But on this console.log we don't expose anything i think that's fine this way.

index.ts
bonustrack: I changed this manually just to test a deployment on Heroku. The initial value was 120, i think 12/10 will not be enough, some page in the UI will need multiple request to the API, and we want to allow script to gather data too, we can start with high amount and see how the hub handle it, then lower if needed.

bonustrack (approved):

PR #64, opened 2021-05-03
Invalid proposals

bonustrack:
**Do not merge this PR**


invalidPropsals.ts
```diff
@@ -0,0 +1,122 @@
...
+const removeInvalidProposals = async () => {
+    try {
+        console.log(`\n\n\nStarted Removing Invalid Proposals from table ${tableName} \n\n\n`);
+        const query = `SELECT * FROM ${tableName} WHERE type = 'proposal'`
+        const data = await db.queryAsync(query, [])
+        console.log(`Total ${data.length} Proposals found`);
+
+        for(const proposal of data) {
+            proposal.payload = jsonParse(proposal.payload)
+            proposal.metadata = jsonParse(proposal.metadata)
```
bonustrack, line 36: Some proposal dont have this field "metadata", it was introduced i think on msg version 0.1.3

invalidPropsals.ts
```diff
@@ -0,0 +1,122 @@
...
+import { jsonParse } from './server/helpers/utils';
+import { spaces } from './server/helpers/spaces';
+import { archiveProposal, loadSpaces } from './server/helpers/adapters/mysql';
+
+const tableName = 'messages';
+const backuptableName = 'messages_backup';
+
+
+const init = async () => {
+    loadSpaces().then(async ensSpaces => {
```
bonustrack, line 16: It's better to take spaces from the hub directly calling `https://hub.snapshot.page/api/spaces`, we would have a more consistent result, sometime loadSpaces fail for some space randomly

invalidPropsals.ts
```diff
@@ -0,0 +1,122 @@
...
+        console.log('ENS spaces', Object.keys(ensSpaces).length);
+
+        await backup();
+        await removeInvalidProposals();
```
bonustrack, line 20: This assume we will execute this function just once, but there is ~7K proposals it will take a while to calculate score for all these. We should have some kind of cursor so we can start the process again where it stoped. Also i think it will i would be better if the verification are done for all proposals of a space at once. Instead of verifying proposal per proposal we can verify space per space that will be much faster.

bonustrack:
@ChaituVR i made a review can you have a look?

ChaituVR:
@bonustrack  added these changes and it is much faster now :)

bonustrack:
@ChaituVR You should run the getScore function only once per space, not once for every proposals, it will make it really much faster.

invalidPropsals.ts
```diff
@@ -0,0 +1,145 @@
...
+            console.log(invalidProposalsInSpace.length + ' proposals marked as invalid');
+        } else {
+            console.log("No proposals replaced in this space");
+        }
+    } catch(error) {
+        console.log(error);
+    }
+}
+
+const backup = async () => {
```
bonustrack, line 133: We don't need this backup in the script. Technicaly the only thing that should change are message type from "proposal" to "invalid-proposal" that should be easy to revert if we see issue. The backup wont be usefull during the process of this script there will be new vote and proposals stored on the table so we wont be able to reuse the backup as it is.

invalidPropsals.ts
```diff
@@ -0,0 +1,145 @@
...
+        return data.json();
+    } catch(error) {
+        console.log(error);
+    }
+}
+
+const markProposalsInvalid = async (invalidProposalsInSpace) => {
+    try {
+        if(invalidProposalsInSpace.length > 0) {
+            const query = `UPDATE ${tableName} SET type = ? WHERE id IN (?)`;
```
bonustrack, line 122: Can you add WHERE type = 'proposal' to make sure we don't change anything else. Also you can use directly `messages` instead of `${tableName}`, we don't change table in this script we don't need to have it dynamic, it will make the script more readable.

ChaituVR:
@bonustrack Updated this branch with those new changes also changed to calculate the score of all proposals in a space at once, please review that

bonustrack:
Can you also put as invalids the proposals from spaces.invalids array?

ChaituVR:
@bonustrack  added

bonustrack:
Just run the script, all good well done! Closing the PR now.

PR #65, opened 2021-05-04
Allow admins to archive any proposal from their space

ChaituVR:
check if admin/owner for https://github.com/snapshot-labs/snapshot/pull/345

### DIFF
--- a/server/helpers/utils.ts
+++ b/server/helpers/utils.ts
@@ -3,8 +3,6 @@ import { verifyMessage } from '@ethersproject/wallet';
 import { convertUtf8ToHex } from '@walletconnect/utils';
 import * as ethUtil from 'ethereumjs-util';
 import { isValidSignature } from './eip1271';
-import { spaces } from './spaces';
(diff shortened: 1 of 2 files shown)


server/helpers/utils.ts
```diff
@@ -105,3 +107,16 @@ export function formatMessage(message) {
...
+export async function isSpaceAdmin(spaceName, address) {
+  if (
+    spaces[spaceName].admins &&
+    spaces[spaceName].admins
+      .map(admin => admin.toLowerCase())
+      .includes(address.toLowerCase())
+  ) {
+    return true;
+  }
+  const spaceUri = await getSpaceUri(spaceName);
```
bonustrack, line 120: We should include the address from the IPNS url here, only the admins.

bonustrack: You dont need a helper function you can call this directly on the verify function

bonustrack (approved):
Looks good!

PR #78, opened 2021-05-17
Use proposal strategies on hub to verify voting power for a vote

ChaituVR:

### DIFF
--- a/server/helpers/adapters/mysql.ts
+++ b/server/helpers/adapters/mysql.ts
@@ -193,8 +193,8 @@ export async function resolveContent(provider, name) {
   return snapshot.utils.decodeContenthash(contentHash);
 }
 
-export async function getProposal(id) {
-  const query = `SELECT * FROM proposals WHERE id = ?`;
-  const proposals = await db.queryAsync(query, [id]);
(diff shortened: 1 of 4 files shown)


ChaituVR:
[Use proposal strategies on hub to verify voting power for a vote](https://trello.com/c/bHSqsX2P/83-use-proposal-strategies-on-hub-to-verify-voting-power-for-a-vote)

server/helpers/adapters/mysql.ts
```diff
@@ -193,8 +193,8 @@ export async function resolveContent(provider, name) {
   return snapshot.utils.decodeContenthash(contentHash);
 }
 
-export async function getProposal(space, id) {
-  const query = `SELECT * FROM messages WHERE space = ? AND id = ? AND type = 'proposal'`;
-  const proposals = await db.queryAsync(query, [space, id]);
+export async function getProposal(id) {
+  const query = `SELECT * FROM proposals WHERE id = ?`;
```
bonustrack, line 197: We should check for the space id here too, we don't want allow someone to delete a proposal from space A, using a message signed with space B. We should also add this check on the archiveProposal function.

PR #83, opened 2021-06-09
Use apollo-server-express instead of graphql-express

ChaituVR:


ChaituVR:
[Use apollo-server-express instead of graphql-express](https://trello.com/c/tNPzGPJA/136-use-apollo-server-express-instead-of-graphql-express)

index.ts
```diff
@@ -33,18 +32,21 @@ app.use(
...
-app.get('/*', (req, res) => res.redirect('/api'));
 
 const PORT = process.env.PORT || 3000;
-app.listen(PORT, () =>
-  console.log(`Snapshot hub started on: http://localhost:${PORT}`)
-);
+const server = new ApolloServer({
+  schema,
+  rootValue,
+  playground: true,
```
bonustrack, line 40: I think this is the default value we can remove it, i got the playground without it using another repo.

bonustrack, line 40: You can add `tracing: true` this will add response time in the returned data

index.ts
```diff
@@ -33,18 +32,21 @@ app.use(
...
+  validationRules: [queryCountLimit(5, 5)]
+});
+server.start().then(() => {
```
bonustrack, line 43: Do we need this? In my code i dont do any start, it look much simpler:
https://github.com/snapshot-labs/highlight/blob/577284dde1006bcbbd0258a26b0afa4c7b6343c5/src/index.ts#L14

index.ts
```diff
@@ -33,18 +32,21 @@ app.use(
...
+  server.applyMiddleware({ app });
+  app.get('/*', (req, res) => res.redirect('/api'));
+  app.listen(PORT, () => {
+    console.log(`Snapshot hub started on: http://localhost:${PORT}`);
+    console.log(
+      `Snapshot graphql api ready at http://localhost:${PORT}${server.graphqlPath}`
```
bonustrack, line 49:
```suggestion
      `GraphQL API ready at http://localhost:${PORT}${server.graphqlPath}`
```

bonustrack:
Can you add "Share" button in the playground too? This should work https://stackoverflow.com/a/60030659

index.ts
ChaituVR: @bonustrack you are able to get this playground on heroku?? 🤔 we need it on production also right? 
because in Heroku `NODE_ENV` will be defaulted to `production`

```
When NODE_ENV is set to production, GraphQL Playground (as well as introspection) is disabled as a production best-practice.
```
https://www.apollographql.com/docs/apollo-server/testing/graphql-playground/

ChaituVR: we need `playground: true,` for production to get it running

bonustrack: Ho yes good catch

index.ts
```diff
@@ -33,18 +32,24 @@ app.use(
...
-  graphqlHTTP({
-    schema,
-    rootValue,
-    graphiql: { defaultQuery },
-    validationRules: [queryCountLimit(5, 5)]
-  })
-);
-app.get('/*', (req, res) => res.redirect('/api'));
 
+const server = new ApolloServer({
```
bonustrack, line 36: Can you move this to ./server/graphql too?

bonustrack:
How about the default query from examples.ts, can we get it too? I'm not able to see it.

server/graphql/index.ts
```diff
@@ -1,9 +1,24 @@
+import { ApolloServer } from 'apollo-server-express';
 import path from 'path';
 import fs from 'fs';
 import { makeExecutableSchema } from 'graphql-tools';
+import { queryCountLimit } from './../helpers/utils';
```
bonustrack, line 5: Can you move this to `/graphql/helpers.ts`?

ChaituVR:
> How about the default query from examples.ts, can we get it too? I'm not able to see it.

@bonustrack I have tried this yesterday but
https://github.com/graphql/graphql-playground#usage

The playground can only work for one case from below:
- show default query in tabs 
(or)
- make query autofill from url with `?query=` meaning [example link we gave in docs](https://hub.snapshot.page/graphql?operationName=Proposal&query=query%20Proposal%20%7B%0A%20%20proposal(id%3A%22QmWbpCtwdLzxuLKnMW4Vv4MPFd2pdPX71YBKPasfZxqLUS%22)%20%7B%0A%20%20%20%20id%0A%20%20%20%20title%0A%20%20%20%20body%0A%20%20%20%20choices%0A%20%20%20%20start%0A%20%20%20%20end%0A%20%20%20%20snapshot%0A%20%20%20%20state%0A%20%20%20%20author%0A%20%20%20%20space%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%20%20name%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D) may not work.. not sure how much time graphqlbin can store our sharable link.. so 🤔 
and also `Note: When using this feature, tabs will be resetted each time the page is reloaded`  so if we write a query and reload our default query will be shown

bonustrack:
Can you try using default query in tabs?

bonustrack:
Closing this PR, there is too much remaining concerns with Apollo playground to worth the change for now.

PR #111, opened 2021-09-17
Votes one day count to explore

ChaituVR:


ChaituVR:
[Add field "1d_votes"](https://trello.com/c/Lg9CjSz6/331-add-field-1dvotes)

bonustrack:
@ChaituVR Can you also get total votes since space creation?

src/helpers/adapters/mysql.ts
```diff
@@ -186,6 +186,14 @@ export async function getFollowers() {
   return await db.queryAsync(query);
 }
 
+export async function getOneDayVotes() {
+  const query = `
+    SELECT space, count(DISTINCT voter, proposal) as count FROM votes
```
bonustrack, line 191: Can you make the SQL syntax in uppercase like this:
```suggestion
    SELECT space, COUNT(DISTINCT voter, proposal) AS count FROM votes
```

src/helpers/adapters/mysql.ts
```diff
@@ -186,6 +186,14 @@ export async function getFollowers() {
...
+    where created > (UNIX_TIMESTAMP() - 86400) group by space
```
bonustrack, line 192:
```suggestion
    WHERE created > (UNIX_TIMESTAMP() - 86400) GROUP BY space
```

src/helpers/spaces.ts
```diff
@@ -26,6 +31,13 @@ setInterval(() => {
       }
     })
   );
+  getOneDayVotes().then((result: any) =>
```
bonustrack, line 34: There should be only 1 "setInterval" that run all the functions

ChaituVR:
Getting unique votes for the last one day works, getting all votes count (nonunique) works, but SQL is not able to get unique + all votes (i takes so much time and fails) tried many ways but no luck,
if this metric is not that important and just for sorting we can use nonunique
`unique =  votes having same proposal and same voter`
or any other ideas that can help?

ChaituVR:
This is not needed anymore for #127

PR #112, opened 2021-09-17
Fix votes query

### LINKED ISSUE #118: Fix Votes query

ChaituVR:
Fixes #118


ChaituVR:
[Query for votes fail on certain queries](https://trello.com/c/mvuVWnAK/295-query-for-votes-fail-on-certain-queries)

ChaituVR:
@bonustrack here without `spaces.settings` everything just works fine, we get the space data anyway because space is a foriegn key

bonustrack:
Make sure it doesnt break space loading for vote

src/graphql/operations/votes.ts
```diff
@@ -47,6 +51,28 @@ export default async function(parent, args, context, info) {
     return Promise.reject('request failed');
   }
 
+  if (requestedFields.space && votes.length > 0) {
+    const spaceIds = votes.map(vote => vote.space.id);
+    const query = `
```
midgerate, line 56: Have we thought about using knex for building SQL query? Its a bit more declarative. Though not requires. since we don't do too many db queries anyway.

src/graphql/operations/votes.ts
```diff
@@ -47,6 +51,28 @@ export default async function(parent, args, context, info) {
...
+      SELECT id, settings FROM spaces
+      WHERE id IN (?) AND settings IS NOT NULL  
+    `;
+    try {
+      let spaces = await db.queryAsync(query, [spaceIds]);
+
+      spaces = Object.fromEntries(
+        spaces.map(space => [space.id, formatSpace(space.id, space.settings)])
+      );
+      votes = votes.map(vote => {
```
midgerate, line 66: if you want to mutate vote, then you don't need a map, you can just use forEach. 

If you want to use map, I would recommended that you do not mutate vote. 
```js
// pseudocode
votes.map(vote => {
  if(condition) return {...vote, space: spaces[vote.space.id]}
  return vote
}
```

ChaituVR: @midgerate Changed vote to not to mutate

ChaituVR:
@bonustrack Please feel free to merge this :P

src/graphql/operations/votes.ts
```diff
@@ -47,6 +51,29 @@ export default async function(parent, args, context, info) {
     return Promise.reject('request failed');
   }
 
+  if (requestedFields.space && votes.length > 0) {
+    const spaceIds = votes.map(vote => vote.space.id);
```
bonustrack, line 55: Can you make that the array have only unique values?

bonustrack (approved):

PR #133, opened 2021-10-06
remove .env from repo

mktcode:
Should not be in repo to avoid accidentally committing credentials, like in db string.


mktcode:
Could also add a `"postinstall": "cp -n .env.example .env"` to package.json.

(-n to not overwrite existing .env)

samuveth:
Deployment seems to be stuck?

mktcode:
> Deployment seems to be stuck?

Deployment needs to be adjusted I guess. It won't automatically copy the .env.example. That's why I suggest putting that into `postinstall`.

midgerate:
@mktcode No, I dont think that's a good idea, the env variables needs to be configured to Fleek.

mktcode:
> @mktcode No, I dont think that's a good idea, the env variables needs to be configured to Fleek.

Hmm... then I don't understand why it even breaks the deployment, if fleek uses it's own configuration.

PR #137, opened 2021-10-07
Fix build issue - '@ensdomains/address-encoder@0.1.9' is not in the npm registery

ChaituVR:
@ensdomains/ensjs is not used anymore and its sub dependency @ensdomains/address-encoder is removed from npm so uninstalled ensjs from hub


ChaituVR:
but works fine in local 😴 Lets wait ⌚

PR #147, opened 2021-10-20
Reduce code

aiinkiestism:
Fixes #853 (https://github.com/snapshot-labs/snapshot/issues/853) .

- Reduce code, remove code duplicate, simplify things, remove unused dependency

I'm also working on these at the same time.
https://github.com/snapshot-labs/snapshot/pull/912
https://github.com/snapshot-labs/snapshot.js/pull/400


package.json
```diff
@@ -9,20 +9,13 @@
...
-    "@ethersproject/hash": "^5.4.0",
-    "@ethersproject/providers": "^5.3.1",
-    "@ethersproject/units": "^5.0.1",
     "@ethersproject/wallet": "^5.0.1",
     "@fleekhq/fleek-storage-js": "^1.0.15",
     "@pinata/sdk": "^1.1.14",
     "@pusher/push-notifications-server": "^1.2.4",
     "@snapshot-labs/snapshot.js": "^0.2.19",
     "@walletconnect/utils": "^1.3.1",
-    "ajv-formats": "^2.0.2",
```
mktcode, line 25: Seems these dependencies are still in package-lock.json....

mktcode (changes requested):
`package-lock.json` needs to be updated.

package.json
```diff
@@ -9,20 +9,13 @@
...
-    "@ethersproject/constants": "^5.0.1",
-    "@ethersproject/contracts": "^5.0.1",
-    "@ethersproject/hash": "^5.4.0",
-    "@ethersproject/providers": "^5.3.1",
-    "@ethersproject/units": "^5.0.1",
     "@ethersproject/wallet": "^5.0.1",
     "@fleekhq/fleek-storage-js": "^1.0.15",
     "@pinata/sdk": "^1.1.14",
     "@pusher/push-notifications-server": "^1.2.4",
     "@snapshot-labs/snapshot.js": "^0.2.19",
```
mktcode, line 17: We'll have to update `snapshot.js` for the changes to work.

mktcode (changes requested):
Snapshot.js version number needs to be upped once this is published: https://github.com/snapshot-labs/snapshot.js/pull/400

package.json
aiinkiestism: Which version should I change to? 0.2.20?

src/writer/settings.ts
```diff
@@ -3,7 +3,7 @@ import isEqual from 'lodash/isEqual';
 import { storeSettings } from '../helpers/adapters/mysql';
 import { jsonParse } from '../helpers/utils';
 import { spaces } from '../helpers/spaces';
-import { getSpaceUri } from '../helpers/ens';
+import { getSpaceUri } from '@snapshot-labs/snapshot.js/src/utils';
```
mktcode, line 6: This can be removed since we already import the entire lib in line 1. Usage needs to be adjusted further down.

```suggestion
```
and
```js
const spaceUri = await getSpaceUri(msg.space);
```
would need to be
```js
const spaceUri = await snapshot.utils.getSpaceUri(msg.space);
```

mktcode (changes requested):
Change needed here as well: https://github.com/aiinkiestism/snapshot-hub/blob/7f111332b0f218b8a09dff236303bdac9eed181d/src/writer/vote.ts#L50-L57

Also snapshot.js version number needs to be increased.

ChaituVR:
@aiinkiestism  would be great if you can remove this line, we want to include this change in this PR else it would break stuff :)
> This line can also be removed:
> https://github.com/aiinkiestism/snapshot-hub/blob/7f111332b0f218b8a09dff236303bdac9eed181d/src/writer/vote.ts#L54

ChaituVR (approved):
LGTM 👍

mktcode (approved):
Looks good. :)

src/helpers/ens.ts
```diff
@@ -1,7 +1,6 @@
 import fetch from 'cross-fetch';
-import { namehash } from '@ethersproject/hash';
-import snapshot from '@snapshot-labs/snapshot.js';
 import gateways from '@snapshot-labs/snapshot.js/src/gateways.json';
+import { getSpaceUri } from '@snapshot-labs/snapshot.js/src/utils';
```
bonustrack, line 3: We shouldnt import function from snapshot.js src folder, can you use snapshot.utils.getSpaceUri instead?

bonustrack:
@aiinkiestism The build fail with this error:
image
You must use "lockfileVersion": 1 on package-lock, not version 2

aiinkiestism:
Is this change OK? Sorry I don't know much about that error because I don't know the error due to package-lock...

src/helpers/ens.ts
mktcode: It's like that in many places. I think you could also set the entry point differently or so, to get rid of the `src` part. And isn't the treeshaking approach the perferred one over importing the entire library?

bonustrack:
@aiinkiestism No, it's not ok, you are still using package lock version 2. You don't get v1 by updating the version like this https://github.com/snapshot-labs/snapshot-hub/pull/147/commits/e0264eb827b0dbe557c42359d584f5753d66765b The syntax is different, you need to use a previous version of NPM, I'm using v6.14.15.

aiinkiestism:
@bonustrack 
Thanks for teaching that! Just updated package-lock with v6.14.15 npm.

PR #149, opened 2021-10-22
feat(): send push notifications when proposal is created

midgerate:
It needs the Pusher Beams keys to be configured in the env variables


src/events/index.ts
```diff
@@ -20,8 +62,12 @@ async function processEvents() {
   const events = await db.queryAsync('SELECT * FROM events WHERE expire <= ?', [
     ts
   ]);
+
   console.log('Process event start', ts, events.length);
+
   for (const event of events) {
+    if (event.event === 'proposal/start') sendPushNotification(event);
```
bonustrack, line 69: There should be a way to check if "SERVICE_PUSH_NOTIFICATIONS"  is enabled like this:
```const servicePushNotifications = parseInt(process.env.SERVICE_PUSH_NOTIFICATIONS || '0');```
```suggestion
    if (servicePushNotifications && event.event === 'proposal/start') sendPushNotification(event);
```

.env.example
```diff
@@ -9,3 +9,6 @@ PINATA_API_KEY=123...
 PINATA_SECRET_API_KEY=123...
 SERVICE_EVENTS=0
 SERVICE_ARWEAVE=0
+PUSHER_BEAMS_INSTANCE_ID=
+PUSHER_BEAMS_SECRET_KEY=
```
bonustrack, line 13: Can you add a prefix `SERVICE_` to these vars? SERVICE prefix mean the feature is optional. 
```suggestion
SERVICE_PUSH_NOTIFICATIONS=0
SERVICE_PUSHER_BEAMS_INSTANCE_ID=
SERVICE_PUSHER_BEAMS_SECRET_KEY=
```

bonustrack (approved):
Looks good!

src/events/index.ts
```diff
@@ -1,10 +1,55 @@
...
+const getSubscribedWallets = async space => {
+  const subscriptions = await db.queryAsync(
+    'SELECT * FROM subscriptions WHERE space = ?',
+    [space]
+  );
+
+  const wallets = subscriptions.map(subscription => subscription.address);
+  return wallets;
+};
+const sendPushNotification = async event => {
```
bonustrack, line 36:
```suggestion

const sendPushNotification = async event => {
```

PR #156, opened 2021-10-30
Fix quadratic voting for negative values

### LINKED ISSUE #153: Enforce positive values in choices object for weighted and quadratic voting
It should only be possible to vote with a value greater than zero in the choice object. Sending a vote like this f.e `{ "1": -1 }` should throw an error.

ChaituVR:
Fixes #153


samuveth (approved):
Thanks!

PR #162, opened 2021-11-02
Snapshot.js 0.3.0

mktcode:
According to: https://github.com/snapshot-labs/snapshot/pull/995

### DIFF
--- a/src/writer/vote.ts
+++ b/src/writer/vote.ts
@@ -51,7 +51,6 @@ export async function verify(body): Promise<any> {
       msg.space,
       jsonParse(proposal.strategies),
       proposal.network,
-      snapshot.utils.getProvider(proposal.network),
       [body.address],
       proposal.snapshot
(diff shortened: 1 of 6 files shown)


bonustrack:
@mktcode How is this different from #147?

mktcode:
@bonustrack 🤦🏻‍♂️ My fault... the other one needs some changes though, but yea totally forgot there's already a PR. :D

src/writer/vote.ts
```diff
@@ -51,7 +51,6 @@ export async function verify(body): Promise<any> {
       msg.space,
       jsonParse(proposal.strategies),
       proposal.network,
-      snapshot.utils.getProvider(proposal.network),
```
ChaituVR, line 54: We still need this change, which is not done in https://github.com/snapshot-labs/snapshot-hub/pull/147

PR #179, opened 2021-11-12
added unique votes db index and adjusted insert query

### LINKED ISSUE #170: Don't store duplicate vote
When someone voting multiple time we store all the votes, instead we should update latest vote and make sure we have only latest vote cast for an address

mktcode:
Fixes #170

**Can be merged once duplicates are removed.**

Making the combination of `voter` and `proposal` a unique key on the `votes` table.

After removing the duplicates, this can be applied:

`ALTER TABLE votes ADD UNIQUE unique_vote(voter, proposal);`

This allows to update existing votes with a simple adjustment to the [`INSERT` query](https://github.com/snapshot-labs/snapshot-hub/pull/179/files#diff-c37d8dbcf5fca29ac1702275af0ecd6ccd2898251cf365e30ff166d88c795fb3R158) (also applies to primary key).

The [votes query](https://github.com/snapshot-labs/snapshot-hub/pull/179/files#diff-5df19be1e5961b913987a2dc3ded98aa1e6cb684ab641cb227469565dd55e61dR51) can then be simplified.


bonustrack:
This is not needed anymore

PR #183, opened 2021-11-13
fix(): only store the current vote

midgerate:


src/helpers/adapters/mysql.ts
```diff
@@ -155,7 +155,20 @@ export async function storeVote(space, body, ipfs, receipt, id) {
     vp_state: ''
   };
 
-  await db.queryAsync('INSERT IGNORE INTO votes SET ?', params);
+  const connection = await db.getConnectionAsync();
+
+  await connection.beginTransactionAsync();
```
bonustrack, line 160: Hum I think transaction is not ideal cuz between start and commit the table canot be updated by others queries

src/helpers/adapters/mysql.ts
```diff
@@ -155,7 +155,20 @@ export async function storeVote(space, body, ipfs, receipt, id) {
...
+
+  const deleteVotesQuery =
+    'DELETE FROM votes WHERE voter = ? AND proposal = ?;';
```
bonustrack, line 163: This wont work if someone broadcast a vote with lower timestamp. We will accept lower timestamp as final vote while it should be the one with highest timestamp

PR #191, opened 2021-11-17
Proposal limit

### LINKED ISSUE #165: Allow max 32 proposals per day per space
And max 320 proposals per month

ChaituVR:
Fixes #165

### DIFF
--- a/src/writer/proposal.ts
+++ b/src/writer/proposal.ts
@@ -64,11 +64,11 @@ export async function verify(body): Promise<any> {
       { count_1d: proposalsDayCount, count_30d: proposalsMonthCount }
     ] = await getRecentProposalsCount(space.id);
 
-    if (
-      proposalsDayCount >= proposalDayLimit ||
-      proposalsMonthCount >= proposalMonthLimit
-    ) {
-      return Promise.reject('too many proposals');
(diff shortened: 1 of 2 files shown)


src/writer/proposal.ts
```diff
@@ -52,6 +58,21 @@ export async function verify(body): Promise<any> {
...
+    const [
+      { count_1d: proposalsDayCount, count_30d: proposalsMonthCount }
+    ] = await getRecentProposalsCount(space.id);
+
+    if (
+      proposalsDayCount >= proposalDayLimit ||
+      proposalsMonthCount >= proposalMonthLimit
+    ) {
+      return Promise.reject('too many proposals');
+    }
```
mktcode, line 72: What about like this?

```suggestion
    if (proposalsDayCount >= proposalDayLimit) {
      return Promise.reject('daily proposal limit reached');
    }
    if (proposalsMonthCount >= proposalMonthLimit) {
      return Promise.reject('monthly proposal limit reached');
    }
```

ChaituVR: @bonustrack What do you think?

mktcode (approved):

PR #193, opened 2021-11-19
fix spaces count for strategy and plugin

rootial:
Fixes # https://github.com/snapshot-labs/snapshot/issues/922
## Summary
Strategies might be used multiple times in a space, so as plugins. 
Try to deduplicate strategies and plugins by name before processing it for each space.
## Test

response from  https://snapshot-hub-master-udn6q21tgz.herokuapp.com/api/explore
shows that we have 12 spaces using ticket strategy

image

After checking https://snapshot-hub-master-udn6q21tgz.herokuapp.com/api/spaces 
find there are 12 spaces using that strategy with 1 space use it twice. So the result is now correct.

### DIFF
--- a/src/routes/api.ts
+++ b/src/routes/api.ts
@@ -57,9 +57,10 @@ router.get('/explore', (req, res) => {
       ? networks[space.network] + 1
       : 1;
 
-    space.strategies.forEach(strategy => {
-      strategies[strategy.name] = strategies[strategy.name]
-        ? strategies[strategy.name] + 1
+    const uniqueStrategies = new Set<string>(space.strategies.map((strategy) =>strategy.name));
+    uniqueStrategies.forEach(strategyName => {
+      strategies[strategyName] = strategies[strategyName]
+        ? strategies[strategyName] + 1
         : 1;
     });
 


rootial:
Hi @bonustrack , i just try to fix the space count issue by deduplicating space/plugin by names, i have no db setup in localhost. I'd like to know could you please take a look this pr and help test it locally. thanks

rootial:
@bonustrack @ChaituVR  can you take a look

rootial:
@bonustrack @ChaituVR please take a look , i verify it via deployment.

ChaituVR:
LGTM @bonustrack

src/routes/api.ts
```diff
@@ -57,13 +57,15 @@ router.get('/explore', (req, res) => {
...
+    const uniqueStrategies = new Set<string>(space.strategies.map((strategy) =>strategy.name));
+    uniqueStrategies.forEach(strategyName => {
+      strategies[strategyName] = strategies[strategyName]
+        ? strategies[strategyName] + 1
         : 1;
     });
 
-    Object.keys(space.plugins || {}).forEach(plugin => {
+    const uniquePlugins = new Set(Object.keys(space.plugins || {}));
+    uniquePlugins.forEach(plugin => {
```
mktcode, line 68: I think plugins should be unique per proposal anyway. So this change would not be necessary but doesn't do any harm also of course. Just came to my mind because I'm working on the plugin system and actually wondered for a moment... could there be more than one instance of a plugin in one space? No, not really. That would just mean the plugin is lacking some "have multiple of this or that" functionality.

mktcode (approved):

src/routes/api.ts
rootial: Thanks for feedback, just get rid of this part of change for plugins.

bonustrack (approved):
Works well

PR #194, opened 2021-11-19
Set base path via env var

mktcode:
Allows to run the hub from a sub-directory, e.g. https://mktcode.uber.space/snapshot-hub/api.

(See also: https://github.com/snapshot-labs/snapshot/pull/1084)


ChaituVR (approved):
LGTM

mktcode:
There's also a way to do this with the express router but I think this will also just do it.

PR #198, opened 2021-11-23
fix(): add validation to graphql schema

midgerate:
Fixes issue where the proposal validation is not showing properly in the space settings


src/graphql/schema.gql
```diff
@@ -207,6 +207,7 @@ type Space {
   plugins: Any
   voting: SpaceVoting
   categories: [String]
+  validation: Strategy
```
bonustrack, line 210: We should add a default value on formatSpace otherwise it will break cuz some space dont have this field

midgerate: When I checked the code on snapshot, formatSpace already has a basic Validation set if validation is undefined/ null @bonustrack

bonustrack (approved):

PR #203, opened 2021-11-26
Basic voting "hide abstain" option

mktcode:
Fixes https://github.com/snapshot-labs/snapshot/issues/1120

Snapshot PR: https://github.com/snapshot-labs/snapshot/pull/1124

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -22,8 +22,8 @@ export function formatSpace(id, settings) {
   space.voting.period = space.voting.period || null;
   space.voting.type = space.voting.type || null;
   space.voting.quorum = space.voting.quorum || null;
-  space.voting.blind = space.voting.blind || null;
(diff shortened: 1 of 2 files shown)


src/graphql/helpers.ts
```diff
@@ -23,6 +23,7 @@ export function formatSpace(id, settings) {
   space.voting.type = space.voting.type || null;
   space.voting.quorum = space.voting.quorum || null;
   space.voting.blind = space.voting.blind || null;
+  space.voting.hideAbstain = space.voting.hideAbstain || null;
```
bonustrack, line 26: It should be default to false
```suggestion
  space.voting.hideAbstain = space.voting.hideAbstain || false;
```

bonustrack, line 26: These should return a boolean
```suggestion
  space.voting.blind = space.voting.blind || false;
  space.voting.hideAbstain = space.voting.hideAbstain || false;
```

PR #208, opened 2021-12-03
docker-compose update

stormsson:
The previous docker-compose configuration wouldn't start immediatly and also had port access problems from the host machine

this pull request allows to run docker-compose and access the mysql server from the host machine

## Changes list
- changed snapshot-hub exposed port from 8080 to 3001, to be uniform with .env default example value of 3001
- changed mysql image from mysql to mariadb
- exposed mysql ports to host machine 

@ChaituVR

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,11 +10,11 @@ services:
     depends_on:
       - snapshot-mysql
     ports:
-      - "8080:8080"
+      - "3001:3001"
 
   # Snapshot MySQL instance
-  snapshot-mysql:
-    image: mysql:8.0.21
+  snapshot-mysql:  
+    image: mariadb:10.6
     container_name: snapshot-mysql
     environment:
       MYSQL_USER: "admin"
@@ -23,7 +23,10 @@ services:
       MYSQL_DATABASE: "snapshot"
     volumes:
         - ./server/helpers/database/:/docker-entrypoint-initdb.d
-
+    ports:
+      - "3306:3306"
+      - "33060:33060"
+ 
 networks:
   default:
     name: snapshot-network


docker-compose.yml
```diff
@@ -10,11 +10,11 @@ services:
...
       - snapshot-mysql
     ports:
-      - "8080:8080"
+      - "3001:3001"
 
   # Snapshot MySQL instance
-  snapshot-mysql:
-    image: mysql:8.0.21
+  snapshot-mysql:  
+    image: mariadb:10.6
```
bonustrack, line 17: Why using mariadb here?

stormsson: It could be my mistake but basically I was unable to make it run with mysql: I am unsure if I'm the only one with this issue: I tried 2 different times from scratch and I was never able to even reach the db.
when switching to mariadb it actually worked instantly so i pushed that too 

(obviously if you use this configuration in the production env then I probably wouldn't advise to merge this pr)

---
 
When running `docker-compose build` + up I met several issues
- unble to access mysql from the host machine: this disallows the use of any tool to interact with the db and, more importantly,  import the initial schema (also made me wonder if the db was really up or not) 
- in console this further error appeared;
```
code: 'ER_NOT_SUPPORTED_AUTH_MODE',
snapshot-hub      |   errno: 1251,
snapshot-hub      |   sqlMessage: 'Client does not support authentication protocol requested by server; consider upgrading MySQL client',```

bonustrack:
@stormsson Closing this PR for now, feel free to reopen if you need it. About docker-compose, it was updated recently, you may want to give a new try.

docker-compose.yml
bonustrack: Which version of MySQL were you using?

mktcode: @stormsson Does this maybe solve your mysql issues: https://stackoverflow.com/a/50377944/2560696

PR #209, opened 2021-12-04
FLEEK_BUCKET_NAME env var, used it in mysql adapter. updated .env.example with PINNING_SERVICE also

stormsson:
mysql adapter had an hardcoded bucket name;
added to .env.example FLEEK_BUCKET_NAME var, and used it instead.
the default value is still the hardcoded one to grant retrocompatibility if the value is not compiled

also added to .env.example the var PINNING_SERVICE=pinata, it is already used in the code, but was not present in the .env.example

@bonustrack


src/helpers/adapters/mysql.ts
```diff
@@ -130,7 +130,7 @@ export async function storeSettings(space, body) {
   const result = await fleek.upload({
     apiKey: process.env.FLEEK_API_KEY || '',
     apiSecret: process.env.FLEEK_API_SECRET || '',
-    bucket: 'snapshot-team-bucket',
+    bucket: process.env.FLEEK_BUCKET_NAME || 'snapshot-team-bucket',
```
bonustrack, line 133:
```suggestion
    bucket: process.env.FLEEK_BUCKET_NAME || 'snapshot-team-bucket',
```

.env.example
```diff
@@ -5,8 +5,10 @@ DATABASE_URL=mysql://...
 RELAYER_PK=0x123...
 FLEEK_API_KEY=123...
 FLEEK_API_SECRET=123...
+FLEEK_BUCKET_NAME=bucket-name
```
bonustrack, line 8:
```suggestion
FLEEK_BUCKET_NAME=snapshot-team-bucket
```

.env.example
```diff
@@ -5,8 +5,10 @@ DATABASE_URL=mysql://...
 RELAYER_PK=0x123...
 FLEEK_API_KEY=123...
 FLEEK_API_SECRET=123...
+FLEEK_BUCKET_NAME=snapshot-team-bucket
```
bonustrack, line 8:
```suggestion
FLEEK_BUCKET=snapshot-team-bucket
```

src/helpers/adapters/mysql.ts
bonustrack, line 133:
```suggestion
    bucket: process.env.FLEEK_BUCKET || 'snapshot-team-bucket',
```

bonustrack (approved):
Looks good thanks!

PR #210, opened 2021-12-06
add DEFAULT_NEWORK  in .env.example and use it in settings.ts

stormsson:
`settings.ts` uses `snapshot.utils.getSpaceUri` without the network parameter;

When running on a network different than mainnet  (like rinkeby ) there is no way to pass it here.

- added `DEFAULT_NETWORK` param to .env.example to configure the network id to use (example value = 1 to use mainnet to grant retrocompatibility)
- altered `settings.ts` to fetch & use env var (with fallback value of "1" to grant retrocompatibility ) 

@bonustrack  @mktcode


.env.example
```diff
@@ -13,3 +13,6 @@ SERVICE_PUSH_NOTIFICATIONS=0
 SERVICE_PUSHER_BEAMS_INSTANCE_ID=
 SERVICE_PUSHER_BEAMS_SECRET_KEY=
 SNAPSHOT_URI=http://localhost:3000
+
+#ETH network id to use, as string; "1" = mainnet
+DEFAULT_NETWORK="1"
```
bonustrack, line 18:
```suggestion
DEFAULT_NETWORK=1
```

bonustrack (approved):
Looks good thanks!

mktcode:
@bonustrack @stormsson

`NETWORK`
https://github.com/snapshot-labs/snapshot-hub/blob/fa64304a1905e3edb7dde6d7a044ffb2fc65b998/src/graphql/helpers.ts#L43-L44

`SNAPSHOT_URI`:

https://github.com/snapshot-labs/snapshot-hub/blob/fa64304a1905e3edb7dde6d7a044ffb2fc65b998/src/events/index.ts#L51

and now there's `DEFAULT_NETWORK`:

This is a bit confusing but `NETWORK` could actually be removed and replaced by `SNAPSHOT_URI`. It does nothing more than building the frontend URL for proposal links and/or push notifications.

PR #213, opened 2021-12-07
#166 created updated in graphql

### LINKED ISSUE #166: Return created and updated params for space on the GraphQL API

stormsson:
Fixes #166  .

- the queries are updated to add spaces created_at and updated_at parameters
- the parameters are passed in the `formatSpace` formatter
- all `format*` methods now include the new params if they have an internal `formatSpace` call

@bonustrack the names of the fields are still the original ones in the db, I altered the names in the examples queries in order to avoid mismatch with the db names until they are renamed

cc @mktcode


src/graphql/schema.gql
```diff
@@ -208,6 +208,8 @@ type Space {
   voting: SpaceVoting
   categories: [String]
   validation: Strategy
+  created_at: Int!
+  updated_at: Int!
```
bonustrack, line 212: We should use `created` and `updated`
```suggestion
  created: Int!
  updated: Int!
```

stormsson: @bonustrack updated schema & queries

src/graphql/operations/votes.ts
```diff
@@ -43,7 +43,7 @@ export default async function(parent, args, context?, info?) {
     SELECT v.* FROM votes v
     LEFT OUTER JOIN votes v2 ON
       v.voter = v2.voter AND v.proposal = v2.proposal
-      AND ((v.created < v2.created) OR (v.created = v2.created AND v.id < v2.id))
+      AND ((v.created < v2.created) OR (v.created = v2.created AND v.id < v2.id))
```
mktcode, line 46:
```suggestion
      AND ((v.created < v2.created) OR (v.created = v2.created AND v.id < v2.id))
```

src/graphql/schema.gql
```diff
@@ -208,8 +208,11 @@ type Space {
   voting: SpaceVoting
   categories: [String]
   validation: Strategy
+  created: Int!
+  updated: Int!
 }
 
+
```
mktcode, line 215:
```suggestion
```

src/graphql/helpers.ts
```diff
@@ -1,8 +1,9 @@
 import { jsonParse } from '../helpers/utils';
+import space from './operations/space';
```
mktcode, line 2: This import is not used. Added automatically by some auto-completion feature in your IDE or so maybe?

```suggestion
```

mktcode: Rest looks good btw.

src/graphql/operations/vote.ts
```diff
@@ -3,7 +3,7 @@ import { formatVote } from '../helpers';
 
 export default async function(parent, { id }) {
   const query = `
-    SELECT v.*, spaces.settings FROM votes v
+    SELECT v.*, spaces.settings, spaces.created_at, spaces.updated_at FROM votes v
```
mktcode, line 6: The column aliases (`space_created_at`, `space_updated_at`) are missing here but are expected in `formatVote()`.

mktcode (changes requested):

src/graphql/helpers.ts
stormsson: @mktcode I believe that your hypothesis is correct and I didn't notice...

src/graphql/operations/vote.ts
stormsson: Uhm I'm a bit unsure about what's happening here, I see them present in my repo branch...  

https://github.com/stormsson/snapshot-hub/blob/%23166-created-updated-in-graphql/src/graphql/operations/vote.ts

mktcode (approved):
Tested locally. Works for me now. :)

src/graphql/operations/votes.ts
```diff
@@ -83,7 +91,7 @@ export default async function(parent, args, context?, info?) {
   if (requestedFields.proposal && votes.length > 0) {
     const proposalIds = votes.map(vote => vote.proposal);
     const query = `
-      SELECT p.*, spaces.settings FROM proposals p
+      SELECT p.*, spaces.settings, spaces.created_at as space_created_at, spaces.updated_at as space_updated_at FROM proposals p
```
bonustrack, line 94: Can you use uppercase for MySQL commands and remove _at if you can like this
```suggestion
      SELECT p.*, spaces.settings, spaces.created_at AS space_created, spaces.updated_at AS space_updated FROM proposals p
```

stormsson: integrated changes in all files

mktcode:
@stormsson sorry this one got a bit lost but I'll merge this now into an intermediate branch. We went with another approach for a similar problem, so I'll have to take a look now and consider what's the best way.

PR #231, opened 2022-01-16
Enforce space strategies for proposal

bonustrack:


src/writer/proposal.ts
```diff
@@ -77,5 +85,88 @@ export async function verify(body): Promise<any> {
...
+      receipt
+    }
+  ]);
+
+  /* Store the proposal in dedicated table 'proposals' */
+  const spaceSettings = spaces[space];
+  const author = getAddress(body.address);
+  const created = parseInt(msg.timestamp);
+  const metadata = msg.payload.metadata || {};
+  const strategies = JSON.stringify(spaceSettings.strategies);
```
bonustrack, line 109: The real change is here, before it was:
```ts
const strategies = JSON.stringify(metadata.strategies || spaceSettings.strategies);
```

PR #233, opened 2022-01-18
followerCount for space(s)

mktcode:
Added a `JOIN` to the space queries, to count followers and added a `followerCount` to the Space type.
This affects all GraphQL queries now but I can also make it depend on `followerCount` being present or not. But we need to work on the hub quite a bit anyway so I left it that simple.

Note:
- `Object.assign` to not loose data from the query after using `formatSpace`
- `parent` => `_parent`: Just a naming convention for unused parameters (doesn't belong here, shame on me)

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -208,7 +208,7 @@ type Space {
   voting: SpaceVoting
   categories: [String]
   validation: Strategy
-  followerCount: Int
+  followersCount: Int
 }
 
 type SpaceFilters {
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -208,6 +208,7 @@ type Space {
   voting: SpaceVoting
   categories: [String]
   validation: Strategy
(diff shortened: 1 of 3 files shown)


src/graphql/schema.gql
```diff
@@ -208,6 +208,7 @@ type Space {
   voting: SpaceVoting
   categories: [String]
   validation: Strategy
+  followerCount: Int
```
bonustrack, line 211: I think it should be with a s, like this "followersCount"
```suggestion
  followersCount: Int
```

bonustrack: Looks like what Twitter API is using https://developer.twitter.com/en/docs/twitter-api/v1/data-dictionary/object-model/user

image

mktcode:
renamed here and here https://github.com/snapshot-labs/snapshot/pull/1475

bonustrack (approved):

samuveth (approved):

PR #243, opened 2022-01-28
Expose strategies from Score API

bonustrack:


src/graphql/schema.gql
```diff
@@ -65,7 +65,7 @@ type Query {
 
   plugins: [Item]
 
-  strategies: [Item]
+  strategies: [StrategyItem]
```
mktcode, line 68: Suggestion:

```suggestion
  strategies: [Strategy]
```

src/graphql/schema.gql
```diff
@@ -309,3 +309,13 @@ type Item {
   id: String!
   spacesCount: Int
 }
+
+type StrategyItem {
```
mktcode, line 313:
```suggestion
type Strategy {
```

src/helpers/spaces.ts
```diff
@@ -34,10 +34,10 @@ setInterval(() => {
       }
     })
   );
-}, 30e3);
+}, 32e3);
```
mktcode, line 37: Because we love powers of two? :D

mktcode (commented):
There's a `Strategy` and `StrategyItem` type now. Can't they be one?

PR #248, opened 2022-01-31
Add event emitter caching for votes

bonustrack:


ChaituVR (approved):

ChaituVR:
Why just `votes`? can be applied to all other queries too right?

mktcode:
I assume `ee.ts` stands for (not sure how to name this file but something with) event emitter? But `eventEmitter.ts` didn't feel perfectly right, because it's not really an event emitter but instead uses one to debounce responses? How about `debounceResponse.ts` then? 😃

bonustrack:
@ChaituVR We can apply it to others queries, I just want to test in votes first, I believe that the most heavy queries

bonustrack:
@mktcode Isn't better to use debounce request instead of response? Maybe just "debounce" since it could be used anywhere?

bonustrack:
I'll merge this first anyway

PR #256, opened 2022-02-05
fix: add limit to first arg

bigint:


bonustrack (approved):

ChaituVR (approved):

src/graphql/operations/proposals.ts
```diff
@@ -41,6 +41,7 @@ export default async function(parent, args) {
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
 
   const { first = 20, skip = 0 } = args;
```
bonustrack, line 43: first should use let and not const

bigint: Ahhh i missed it, fixed now ✅

PR #260, opened 2022-02-16
load space metadata right after loading spaces

mktcode:
Currently you have to wait 32 seconds for the explore endpoint to return something, every time you restart the hub, which is a bit annoying during development. So that function is now simply called right after loading the spaces.


ChaituVR (approved):

bonustrack (approved):
Looks good

PR #262, opened 2022-02-21
feat(): add network property to strategy

midgerate:
Added network property which will be later used to display the network in the space settings for each strategy.

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -266,6 +266,7 @@ type Proposal {
 
 type Strategy {
   name: String!
+  network: String
   params: Any
 }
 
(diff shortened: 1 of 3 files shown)


bonustrack:
I think we should auto populate this field with the space network otherwise type validation would fail

src/graphql/schema.gql
```diff
@@ -266,6 +266,7 @@ type Proposal {
 
 type Strategy {
   name: String!
+  network: String
```
samuveth, line 269: Network should be a number I think

samuveth: Maybe not

bonustrack: Yes its a string

bonustrack: Network is not the same as chainId

samuveth (approved):

bonustrack (approved):
Looks good

PR #265, opened 2022-02-22
commented out the currenty not used function

mktcode:
To make Github and my IDE stop complaining:

image

### DIFF
--- a/src/scores/index.ts
+++ b/src/scores/index.ts
@@ -159,18 +159,18 @@ export async function getProposalScores(proposalId) {
   }
 }
 
-async function run() {
-  const [
-    proposal
-  ] = await db.queryAsync(
-    'SELECT id FROM proposals WHERE scores_state = ? ORDER BY created ASC LIMIT 1',
-    ['']
-  );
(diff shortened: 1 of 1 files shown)


ChaituVR:
I am wondering why this line is commented in the first place 🤔 how it is calculating scores for closed proposals now @bonustrack 
```
snapshot.utils.sleep(5000).then(() => run());
````

mktcode:
Yeah I wonder now too... tbh I didn't really care and just saw that it's unused for some reason. But let me guess... on the server it's not commented out? 😃

ChaituVR:
Maybe it is handled in some other place so commented, I don't know :D we need release notes for the hub

ChaituVR (approved):

bonustrack:
@mktcode @ChaituVR This function is used to calculate score on proposals once the server start, I wouldn't comment it. This is something I enable sometime but otherwise the flow is that once a proposal is closed we get the information with events and we trigger score calculation. Here: https://github.com/snapshot-labs/snapshot-hub/blob/master/src/events/index.ts#L76

PR #269, opened 2022-02-23
Network strategy

midgerate:
Adds network to the space strategy.

### DIFF
--- a/src/graphql/examples.ts
+++ b/src/graphql/examples.ts
@@ -13,6 +13,7 @@ query Spaces {
     symbol
     strategies {
       name
+      network
       params
     }
     admins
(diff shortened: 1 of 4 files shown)


mktcode (approved):

midgerate:
So here's the explanation of the problem why strategies is undefined for votes. 

When the votes are loaded, we call `formatVote` which then calls `formatSpace`. 
Now, formatVote passes settings to formatSpace, however, vote object does not contain settings. 
This is because the SQL query used to fetch the votes, does not have a join to space. 

We used to have that earlier, but it was removed due to performance reasons. So adding it back merely wont solve the issue. 
A new issue is created to track this problem. 

https://github.com/snapshot-labs/snapshot-hub/issues/270

samuveth (approved):

PR #277, opened 2022-03-22
Feature - Add Wallet Follow and Unfollow

midgerate:
A new feature to follow and unfollow a particular wallet.


midgerate:
Draft because its blocked by https://github.com/snapshot-labs/snapshot.js/pull/534   
I need to know the correct snapshot.js version.

src/helpers/database/schema.sql
```diff
@@ -122,6 +122,17 @@ CREATE TABLE follows (
   INDEX created (created)
 );
 
+CREATE TABLE walletfollows (
```
bonustrack, line 125: We shouldnt make a new table for it, we should instead use "follows" table, but add a field "type", type can be "space" or "account", later we can also merge "subscription" there.

bonustrack: I'd call it "following"

bonustrack: And add a column "type"

src/ingestor/typedData/index.ts
```diff
@@ -42,14 +42,23 @@ export default async function ingestor(body) {
   let type = hashTypes[hash];
 
   if (
-    !['settings', 'alias'].includes(type) &&
+    !['settings', 'alias', 'walletFollow', 'walletUnfollow'].includes(type) &&
```
samuveth, line 45: Why does walletFollow and walletUnfollow need to added here?

bonustrack:
Closing this for now as it wont be prioritized anytime soon

PR #292, opened 2022-04-07
Sub-DAO fields

mktcode:
The fields `parent` and `children` are implemented as type `Space`. The other option was to make them simple `String` fields, which would have required multiple requests in the UI and any other client (like the mobile app) to fetch additional information about parent or child spaces. This would have resulted in duplicated code across those clients and felt like defying the purpose of a GraphQL API.

Now we can do queries like this:

```
spaces(where: { id_in: $id_in }) {
  id
  name
  parent {
    id
    name
    children {
      id
    }
  }
  children {
    id
    name
    parent {
      id
    }
  }
}
```

The logic to fetch spaces has moved into a helper function `fetchSpaces` and is used in both resolvers (space and spaces). One query is made to fetch the list of requested spaces and one additional query is made in case parent/children data is requested.

**As long as only the id of the children or parent is requested, no extra database query is required.** The `formatSpace` function already returns them in a `parent: { id: 'myspace.eth' }` format.

On the second level (`parent.children` or `children.parent`) only the id can be queried, in order for a client to perform a check, whether a parent-child relationship is defined on both sides.

Queries including `parent.parent` or `children.children` are not allowed at all.


src/graphql/schema.gql
```diff
@@ -222,6 +222,8 @@ type Space {
   validation: Strategy
   followersCount: Int
   proposalsCount: Int
+  parent: Space
```
bonustrack, line 225: Did you gave up on using String?

mktcode: I planned to inform you before marking this ready for review. ;) Gonna add some explanation. Open for discussion and changes. But so far I think it makes sense like this.

src/graphql/operations/spaces.ts
```diff
@@ -1,47 +1,55 @@
...
-  `;
-  try {
-    const spaces = await db.queryAsync(query, params);
-    return spaces.map(space =>
-      Object.assign(space, formatSpace(space.id, space.settings))
-    );
+      // fetch parents and children
+      for (const space of spaces) {
+        if (space.parent?.id) {
+          const parentResult = await fetchSpaces({
```
midgerate, line 36: Could you also do "in" query and get the list of spaces at once without needing to do multiple queries?

midgerate: If it works, then you wont have to calculate and error out when there are more than 100 children.

mktcode: Aaaahhh yea sure that should be a lot better. Collect all ids, make one query and then remap everything in js. Gonna do that.

mktcode: So much better. Not sure why I over-complicated it like that. I also moved the check whether parent and child match, into the resolver now. Less work in clients. Gonna update UI now accordingly.

mktcode:
If I understand correctly you can only apply it to the full schema and then set some ignore rules. It's also not actively maintained it seems. Not sure. Can't we handle this in another issue/PR?

bonustrack:
@mktcode I believe the depth limit should be applied to the whole schema, don't you think so?

mktcode:
yea sure. I installed it now and set it to a max depth of 5, which would still allow queries like this:

```
space (id: "somespace.eth") {
    id
    parent {
      id
      proposals {
        votes {
          user {
            id
            address
            profile {
              bio
              avatar
            }
          }
        }
      }
    }
  }
```

But it's just not really related to the sub-dao feature because we don't even need the full max depth there anyway.

bonustrack:
For some reason when I query for 200 results it fail for me 
image
http://localhost:3000/graphql?query=%0Aquery%20%7B%0A%20%20spaces%20(first%3A%20200)%20%7B%0A%20%20%20%20id%0A%20%20%20%20name%0A%20%20%7D%0A%7D%0A

mktcode:
> For some reason when I query for 200 results it fail for me

Seems this is unrelated to this PR. Happens on master too. Probably some bug in a new version of mysql.

https://stackoverflow.com/questions/67242726/mysql-8-order-by-created-at-out-of-sort-memory-error-1038
https://bugs.mysql.com/bug.php?id=103225

samuveth:
This needs to be reviewed and merged so we can move forward with the frontend part @bonustrack

ChaituVR (approved):

ChaituVR (approved):

bonustrack (approved):

PR #293, opened 2022-04-08
Increase rate limit

mktcode:
"Fixes" https://github.com/snapshot-labs/snapshot/issues/2122

I don't know how the current rate limitation came to be but it seems to be a bit too tight. It's 4 requests per second, which can be achieved manually, e.g. by quickly joining spaces (as in the above issue) or if we'd try to do multiple requests programmatically, as I'd first tried in the Sub-DAO feature to circumvent adjusting the GraphQL API, which I now did instead.

The same issue can also cause the delegation page to break.

image

I just went with doubling the requests now in this PR from 4 to 8 (128 per 16s) but I see that these values have been [adjusted in the past](https://github.com/snapshot-labs/snapshot-hub/commits/master/src/helpers/rateLimit.ts). So I'm not sure if it should be increased like this.

The other aspect of this issue is of course that these things need to be handled properly in the UI. Currently it tends to simply break if there's a problem on the backend side.

### DIFF
--- a/src/helpers/rateLimit.ts
+++ b/src/helpers/rateLimit.ts
@@ -3,7 +3,7 @@ import { sendError, sha256 } from './utils';
 
 export default rateLimit({
   windowMs: 16 * 1e3,
-  max: 64,
+  max: 128,
   handler: (req, res) => {
     const id = sha256(req.ip);
     console.log('Too many requests', id.slice(0, 7));


bonustrack:
I'd prefer if we don't increase this limit, it's not 4 request per second. It's 64 per 16sec, that's quite different. You can send 4 request per second, but you just can't do that forever. If you reach limit and not send 4 request then maybe the issue is on how the hub calculate what is a request. I'd rather look in this instead of increasing the limit. I dont see which user need to do more than 64 request within 16sec

mktcode:
Yea was a bit too quick on this one. Doesn't really make sense to tackle it from this side. Reacting to errors properly in the UI is the better way to go. But that will either be a workaround for this specific case or I'll also have to change snapshot.js. Not sure...

But I'll close this then.

PR #298, opened 2022-04-12
User profile

midgerate:
@samuveth 

https://github.com/snapshot-labs/snapshot.js/pull/558/files


src/graphql/operations/users.ts
```diff
@@ -0,0 +1,44 @@
...
+    created: 'number'
+  };
+  const whereQuery = buildWhereQuery(fields, 'f', where);
+  const queryStr = whereQuery.query;
+  const params: any[] = whereQuery.params;
+
+  let orderBy = args.orderBy || 'created';
+  let orderDirection = args.orderDirection || 'desc';
+  if (!['created'].includes(orderBy)) orderBy = 'created';
+  orderBy = `f.${orderBy}`;
```
mktcode, line 21:
```suggestion
  orderBy = `u.${orderBy}`;
```

mktcode, line 21:
```suggestion
  const whereQuery = buildWhereQuery(fields, 'u', where);
  const queryStr = whereQuery.query;
  const params: any[] = whereQuery.params;

  let orderBy = args.orderBy || 'created';
  let orderDirection = args.orderDirection || 'desc';
  if (!['created'].includes(orderBy)) orderBy = 'created';
  orderBy = `u.${orderBy}`;
```

src/ingestor/typedData/index.ts
```diff
@@ -42,14 +42,18 @@ export default async function ingestor(body) {
   let type = hashTypes[hash];
 
   if (
-    !['settings', 'alias'].includes(type) &&
+    !['settings', 'alias', 'profile'].includes(type) &&
```
mktcode, line 45: Wouldn't it make more sense to check the types that DO need a space property?

src/writer/profile.ts
```diff
@@ -0,0 +1,20 @@
+import { getAddress } from '@ethersproject/address';
+import db from '../helpers/mysql';
+
+export async function verify(): Promise<any> {
+  return true;
```
mktcode, line 5:
```suggestion
  // TODO: implement
  return true;
```
:D

mktcode (changes requested):

midgerate:
Depends on https://github.com/snapshot-labs/snapshot.js/pull/557

src/graphql/schema.gql
```diff
@@ -311,6 +336,16 @@ type Subscription {
   created: Int!
 }
 
+type User {
+  id: String!
+  ipfs: String
+  address: String!
+  username: String
+  bio: String
```
bonustrack, line 344: Sorry missed to review on the snapshot.js PR but the fields should be:

name (instead of username)
about (instead of bio)

Also i think it should be stored as a JSON in the db to get more flexibility if we want add fields on the schema (similar than space table now)

midgerate: so the fields will be profile - JSON

and inside of it name and bio 

is that good?

samuveth: @bonustrack can't this be expressed beforehand?

mktcode: If these are not separate fields in the db table, we should at least use the tools mysql provides for json fields: https://dev.mysql.com/doc/refman/8.0/en/json-search-functions.html#operator_json-column-path
Otherwise this whole design doesn't really make sense to me and just blocks because it's unclear what should be a column and what should go into the json. I mean... we could just have tables with an id, created/updated, ipfs hash, and a json field. But we need to define that somewhere.

midgerate: @mktcode I can use JSON datatype instead of TEXT, however for the purpose of this PR, do you see us using any JSON functions to search / store values?

midgerate:
@mktcode @bonustrack review again please

src/graphql/schema.gql
mktcode: Well, we should actually avoid accessing the JSON data in our SQL queries anyway. So a text field would also be ok. On the other hand, for the custom domains I found that having the domain of a space be an indexed column makes sense but now I'd have to use MySQL JSON functions to do a lookup by domain. We just need to think well about what we need as conditions in SQL queries and what can be just part of the data blob.
Fabien sent me this article: https://www.wix.engineering/post/scaling-to-100m-mysql-is-a-better-nosql
Which helped me a lot to understand the design decisions here.

> Any field that is not used as a condition in a query has been folded into a single blob field

src/ingestor/typedData/index.ts
```diff
@@ -42,14 +42,33 @@ export default async function ingestor(body) {
...
+      'vote',
+      'vote-array',
+      'vote-string',
+      'vote',
+      'vote-array',
+      'vote-string',
+      'follow',
+      'unfollow',
+      'subscribe',
+      'unsubscribe'
```
mktcode, line 59: Some duplicates here.

mktcode (approved):

src/writer/profile.ts
```diff
@@ -3,14 +3,15 @@ import db from '../helpers/mysql';
...
-  }
+  // TODO: add this code once snapshot.js update is released with the profile schema.
+  // const schemaIsValid = snapshot.utils.validateSchema(
+  //   snapshot.schemas.profile,
+  //   body.profile
+  // );
+  // if (schemaIsValid !== true) {
+  //   console.log('[writer] Wrong profile format', schemaIsValid);
+  //   return Promise.reject('wrong profile format');
+  // }
```
mktcode, line 14: I'd make this PR a draft then. Then you also don't need to write capslock titles. 🤪

src/helpers/database/schema.sql
```diff
@@ -145,3 +145,15 @@ CREATE TABLE subscriptions (
   INDEX ipfs (ipfs),
   INDEX created (created)
 );
+
+
+CREATE TABLE users (
+  id VARCHAR(66) NOT NULL,
```
bonustrack, line 151: What's the id? Isn't the same as address? In this case I would not add address here but just id, and use it as primary key

src/ingestor/typedData/index.ts
```diff
@@ -36,20 +36,33 @@ export default async function ingestor(body) {
     return Promise.reject('wrong domain');
 
   const hash = sha256(JSON.stringify(types));
-
   if (!Object.keys(hashTypes).includes(hash))
     return Promise.reject('wrong types');
   let type = hashTypes[hash];
 
   if (
-    !['settings', 'alias'].includes(type) &&
```
bonustrack, line 45: How is it better to do it this way?

src/graphql/schema.gql
```diff
@@ -311,6 +336,17 @@ type Subscription {
...
 }
 
+type User {
+  id: String!
+  ipfs: String
+  address: String!
+  name: String
+  about: String
+  avatar: String
+  twitter: String
```
bonustrack, line 346: We should remove twitter here

src/ingestor/typedData/index.ts
midgerate: Its same just lesser conginitive load to understand. 
image

src/graphql/helpers.ts
```diff
@@ -34,6 +34,15 @@ export function formatSpace(id, settings) {
   return space;
 }
 
+export function formatUser(user) {
+  const profile = jsonParse(user.profile ?? {});
```
bonustrack, line 38: The first argument on jsonParse should be a string, now you are using empty object if profile is empty. You can do like this instead:
```suggestion
  const profile = jsonParse(user.profile, {});
```

src/ingestor/typedData/index.ts
bonustrack: But then you are adding a much bigger list, I'd rather map the action that doesn't require space than the opposite.

bonustrack (approved):
Looks good!

PR #301, opened 2022-04-15
Store vote voting power if no delegation

bonustrack:


src/writer/vote.ts
```diff
@@ -4,6 +4,19 @@ import { jsonParse } from '../helpers/utils';
...
+  const scores = await snapshot.utils.getScores(
+    msg.space,
+    jsonParse(proposal.strategies),
+    proposal.network,
+    [body.address],
+    proposal.snapshot
+  );
+  return scores
+    .map((score: any) => Object.values(score).reduce((a, b: any) => a + b, 0))
+    .reduce((a, b: any) => a + b, 0);
```
mktcode, line 17: Since there's only one address, I find this a bit more readable:

```suggestion
  return scores
    .map((score: any) => Object.values(score)[0])
    .reduce((a, b: any) => a + b, 0);
```

src/writer/vote.ts
```diff
@@ -113,7 +104,31 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
...
+  // Get proposal
+  const proposal = await getProposal(msg.space, msg.payload.proposal);
+
+  // Store voting power if no delegation strategies
+  let withDelegation = false;
+  const strategies = jsonParse(proposal.strategies);
+  const names = strategies.map(strategy => strategy.name);
+  names.forEach(name => {
+    if (name.includes('delegation')) withDelegation = true;
+  });
```
mktcode, line 116: A bit shorter:

```suggestion
  const strategies = jsonParse(proposal.strategies);
  const withDelegation = strategies.some(strategy => strategy.name.includes('delegation'));
```

src/writer/vote.ts
```diff
@@ -132,6 +147,21 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
...
+    ipfs,
+    voter,
+    created: parseInt(msg.timestamp),
+    space: msg.space,
+    proposal: msg.payload.proposal,
+    choice: JSON.stringify(msg.payload.choice),
+    metadata: JSON.stringify(msg.payload.metadata || {}),
+    vp,
+    vp_by_strategy: JSON.stringify([]),
+    vp_state: '',
```
mktcode, line 161: These can't also be set right away?

mktcode (approved):
Made suggestions but can't find issues. Approved.

src/writer/vote.ts
bonustrack: Yes i'm doing it on another branch but its still a draft

src/writer/vote.ts
```diff
@@ -113,7 +104,31 @@ export async function action(body, ipfs, receipt, id): Promise<void> {
...
-  //
+  // Get proposal
+  const proposal = await getProposal(msg.space, msg.payload.proposal);
+
+  // Store voting power if no delegation strategies
+  let withDelegation = false;
+  const strategies = jsonParse(proposal.strategies);
+  const names = strategies.map(strategy => strategy.name);
+  names.forEach(name => {
+    if (name.includes('delegation')) withDelegation = true;
```
ChaituVR, line 115: Should we also add `erc20-with-delegation`?

bonustrack: This already check for erc20-with-delegation

mktcode: @ChaituVR anything that `includes` `'delegations'.`

ChaituVR: Oh!!! got it. how did I miss that.. thanks, @mktcode :D

mktcode: Maybe the code should be [shortened a bit](https://github.com/snapshot-labs/snapshot-hub/pull/301#discussion_r851613946). :D

PR #302, opened 2022-04-16
enforce min value 1 for choices in basic and single-choice

mktcode:
Choices starting at 1 was not enforced in the hub, leaving the possibility of a custom integration sending 0 and also calculating and displaying results accordingly.


bonustrack (approved):

ChaituVR (approved):
Looks good , not related to this but in different PR can we move "Invalid choice" check to a separate function, right now we have three places that can return `invalid choice', 
should be like
```js
if(checkInvalidChoice(payload, proposal)){
  return Promise.reject('invalid choice');
}
```

PR #310, opened 2022-04-29
feat: filter votes by choice

### LINKED ISSUE #290: [GraphQL]  Filter votes by choice
Currently there is no way to filter votes by choice. There should be a where-filter taking an array of choices to return.

mktcode:
See https://github.com/snapshot-labs/snapshot-hub/issues/290#issuecomment-1113165105

but this breaks pagination :(

### DIFF
--- a/src/graphql/operations/votes.ts
+++ b/src/graphql/operations/votes.ts
@@ -56,6 +56,39 @@ async function query(parent, args, context?, info?) {
     return Promise.reject('request failed');
   }
 
+  // filter by choice
+  if (where.choice) {
+    votes = votes.filter(v => {
+      if (typeof v.choice === 'number') {
+        return v.choice === where.choice;
+      }
+      if (typeof v.choice === 'object') {
+        if (Array.isArray(v.choice)) {
+          return v.choice.includes(where.choice);
+        } else {
+          return Object.keys(v.choice).includes(where.choice.toString());
+        }
+      }
+      return false;
+    });
+  } else if (where.choice_in) {
+    votes = votes.filter(v => {
+      if (typeof v.choice === 'number') {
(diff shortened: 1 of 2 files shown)


src/graphql/operations/votes.ts
```diff
@@ -56,6 +56,39 @@ async function query(parent, args, context?, info?) {
...
+      }
+      if (typeof v.choice === 'object') {
+        if (Array.isArray(v.choice)) {
+          return v.choice.includes(where.choice);
+        } else {
+          return Object.keys(v.choice).includes(where.choice.toString());
+        }
+      }
+      return false;
+    });
```
ChaituVR, line 73: This may affect `first` and `skip` for pagination right?

ChaituVR: Maybe we should have a new input variable `filter` instead of `where`

ChaituVR (changes requested):

src/graphql/operations/votes.ts
mktcode: You mean as a clear way of saying "this filtering happens after pagination"?

For a moment I thought about doing the pagination in JS too. Removing the `LIMIT` from the query, at least when there's a choice filter. But that doesn't feel right.

So yea... idk. We just tell people it's not possible and they have to do the filtering themselves, in their client?

bonustrack: @mktcode Yes, they can filter on their side, we would implement this when we have an elegant way to do it. Who is asking this feature?

bonustrack:
Closing this PR for now, it's unlikely we will allow query for specific choice as the choice field is a JSON and query into doesn't scale well.

PR #317, opened 2022-05-13
feat(): allow native push notifications to work

midgerate:
Needed for https://github.com/snapshot-labs/snapshot-app/issues/14


bonustrack:
This is not relevant anymore (should be on snapshot-webhook)

PR #320, opened 2022-05-19
Added domain column, DNS check and GraphQL argument

mktcode:
As part of: https://github.com/snapshot-labs/snapshot/issues/2203

When storing settings, containing a custom domain, it's CNAME will be checked and the domain will be stored in its own column. It will also still be part of the settings JSON.

The `space` graphql query accepts a domain instead of id.

These changes should not break anything, as long as the column/index are added in the DB. But it won't be possible to save settings with a domain not having the correct CNAME.


ChaituVR (approved):

mktcode:
@bonustrack @ChaituVR I just resolved the conflicts here. Not sure how we want to proceed with this and who. I'll focus on Safesnap this month.

mktcode:
Keeping this branch in sync, in case it's still relevant.

ChaituVR:
Closing this as stale

PR #321, opened 2022-05-20
Fix strategies null or object

ChaituVR:
Problem:
Try this following query on hub:
```JS
query {
    proposals (
    first: 62
    skip: 0
    orderBy: "created"
    orderDirection: asc
    where: {
      created_gt: 1637909052
      state: "closed"
    }
  ) {
    id
  }
 }
 ```
 
 We get this error:
 
 ```js
 ﻿[graphql] TypeError: proposal.strategies.map is not a function﻿
```

Reason: previously we used to accept whatever the metadata sent from the proposal payload, so I assume there could be an object in `strategies` instead of an array, this PR will handle that case and will return an empty array if `strategies` is not an array


bonustrack (approved):

ChaituVR:
Not needed anymore

Issue #351, opened 2022-06-27
schema.sql in readme.md 404s


i001962:
### Expected behavior
Clicking link in readme should 200
### Actual behavior
Clicking link in readme 404s
### Steps to reproduce the behavior
Load this https://github.com/snapshot-labs/snapshot-hub
then click this https://github.com/snapshot-labs/snapshot-hub/blob/master/src/helpers/database/schema.sql
404s

mktcode:
👍🏻

PR #352, opened 2022-06-27
Update README.md

i001962:
correct path - https://github.com/snapshot-labs/snapshot-hub/blob/master/src/helpers/schema.sql

Fixes # .https://github.com/snapshot-labs/snapshot-hub/issues/351

- 
- 
-


bonustrack (approved):
Good catch thanks for the PR @i001962!

PR #381, opened 2022-07-25
Add GraphQL operation to get vp

bonustrack:


src/helpers/strategies.ts
```diff
@@ -35,6 +35,14 @@ async function loadStrategies() {
   );
 }
 
-setInterval(async () => await loadStrategies(), 60e3);
+async function run() {
+  try {
+    await loadStrategies();
+  } catch (e) {
+    console.log('[spaces] Failed to load', e);
```
ChaituVR, line 42:
```suggestion
    console.log('[strategies] Failed to load', e);
```

src/helpers/strategies.ts
```diff
@@ -35,6 +35,14 @@ async function loadStrategies() {
...
 
-setInterval(async () => await loadStrategies(), 60e3);
+async function run() {
+  try {
+    await loadStrategies();
+  } catch (e) {
+    console.log('[strategies] Failed to load', e);
+  }
+  await snapshot.utils.sleep(60e3);
+  await run();
```
ChaituVR, line 45: This function is never-ending
```suggestion
  run();
```

PR #458, opened 2022-09-16
Store voting power when cast vote

bonustrack:


src/ingestor/writer/vote.ts
```diff
@@ -57,14 +57,25 @@ export async function verify(body): Promise<any> {
...
   const metadata = JSON.stringify(msg.payload.metadata || {});
   const app = msg.payload.app || '';
   const reason = msg.payload.reason || '';
+
+  // Check if voting power is final
+  let withDelegation = false;
+  context.proposal.strategies
+    .map(strategy => strategy.name)
+    .forEach(name => {
+      if (name.includes('delegation')) withDelegation = true;
```
ChaituVR, line 74: Everything looks great 🚀 except that delegation can be used in child strategy too, recently POH used math strategy on top of delegation, maybe there can be grand child strategy too :sweat_smile: (we should limit these levels of strategy) 

I think best solution for now would be to stringify strategies object and check if there is delegation string

For delegation strategy, would be great if there is a way to just update delegate's VP (when delegator vote)

ChaituVR (changes requested):

PR #467, opened 2022-09-25
Optimize proposal scores and votes vp update

bonustrack:
This PR optimize the function `getProposalScores` (which is running in loop to update proposals scores and votes). This includes:

1. Update only votes on the db when there is changes in voting power or state.
2. Don't use getScores if all votes are final.


ChaituVR:
Should also merge this:
- [ ] https://github.com/snapshot-labs/snapshot-webhook/pull/33

src/scores.ts
```diff
@@ -1,8 +1,9 @@
 import fetch from 'cross-fetch';
 import snapshot from '@snapshot-labs/snapshot.js';
 import db from './helpers/mysql';
+import { sha256 } from './helpers/utils';
 
-async function getProposal(id) {
+async function getProposal(id): Promise<any | undefined> {
```
ChaituVR, line 6:
```suggestion
async function getProposal(id: string): Promise<any | undefined> {
```

src/scores.ts
```diff
@@ -62,127 +67,135 @@ export async function getScores(
   }
 }
 
+async function updateVotesVp(votes: any[], vpState: string, proposalId: string) {
+  const votesWithChange = votes.filter(vote => {
+    const key1 = sha256(JSON.stringify([vote.balance, vote.scores, vpState]));
+    const key2 = sha256(JSON.stringify([vote.vp, vote.vp_by_strategy, vote.vp_state]));
+    return key1 !== key2;
```
ChaituVR, line 74: Why don't we just compare strings directly? we are not reusing these keys, right? I feel it is better to compare strings directly

src/scores.ts
```diff
@@ -62,127 +67,135 @@ export async function getScores(
...
+  });
+  if (votesWithChange.length === 0) return;
+
+  const max = 200;
+  const pages = Math.ceil(votesWithChange.length / max);
+  const votesInPages: any = [];
+  Array.from(Array(pages)).forEach((x, i) => {
+    votesInPages.push(votesWithChange.slice(max * i, max * (i + 1)));
+  });
```
ChaituVR, line 83: Maybe better to use this, 
```suggestion
  for (let i = 0; i < pages; i++) {
    votesInPages.push(votesWithChange.slice(max * i, max * (i + 1)));
  }
```

src/scores.ts
```diff
@@ -62,127 +67,135 @@ export async function getScores(
...
-    );
-
-    // Add vp to votes
-    votes = votes.map((vote: any) => {
-      vote.scores = proposal.strategies.map((strategy, i) => scores[i][vote.voter] || 0);
-      vote.balance = vote.scores.reduce((a, b: any) => a + b, 0);
-      return vote;
-    });
+    let isFinal = true;
+    votes.forEach(vote => (isFinal = vote.vp_state !== 'final' ? false : isFinal));
```
ChaituVR, line 138:
```suggestion
    isFinal = votes.every(vote => vote.vp_state === 'final');
```

ChaituVR (approved):

src/scores.ts
bonustrack: They are not only string, scores by strategy is an array

Issue #413, opened 2022-08-09
Don't return error object in response


ChaituVR:
Return a valid error

bonustrack:
Need to reproduce

zzuziak:
Hey hello @ChaituVR what's the status here? 🙏

PR #478, opened 2022-10-10
Error message instead of object

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot-hub/issues/413


wa0x6e:
I think this can be closed, as most of the files does not exist anymore

PR #484, opened 2022-10-17
Update hasStrategyOverride

ChaituVR:


src/helpers/utils.ts
```diff
@@ -47,11 +47,15 @@ export function rpcError(res, code, e, id) {
 }
 
 export function hasStrategyOverride(strategies: any[]) {
-  const keywords = [
+  const keywords =  [
+    'api',
```
ChaituVR, line 51:
```suggestion
    '"api"',
    'api-post',
```

src/helpers/utils.ts
```diff
@@ -47,11 +47,22 @@ export function rpcError(res, code, e, id) {
...
-    'aura-balance-of-vlaura-vebal'
+  const keywords =  [
+    '"api"',
+    '"api-post"',
+    '"aura-vlaura-vebal-with-overrides"',
+    '"balance-of-with-linear-vesting-power"',
+    '"balancer-delegation"',
+    '"cyberkongz"',
+    '"cyberkongz-v2"',
+    '"delegation"',
```
bonustrack, line 58: Would this be safer or wont change anything?
```suggestion
    'delegation',
```

ChaituVR: Yep, some strategies with `delegation` name don't actually depend on other `addresses`

I just check them again:

- `erc20-balance-of-delegation` - In this list
- `erc20-balance-of-quadratic-delegation` - In this list
- `esd-delegation` - In this list
- `xdai-stake-delegation` - Doesn't depend, It checks the voting power of delegators and return it, no override
- `delegation` - In this list
- `the-graph-delegation` - They just pass delegation keyword to check if it is a valid strategy or not
- `celer-sgn-delegation` - They get the VP of delegations, no override
- `balancer-delegation` - In this list
- 'meebitsdao-delegation` - No override
- `orbs-network-delegation` - In this list
- `nation3-votes-with-delegations` - No override

PR #488, opened 2022-10-19
Move authors and onlyMembers validation to the hub

bonustrack:
Currently we do this check on Snapshot.js, with this PR I'm moving the logic to allow authors to publish a proposal as a core feature.

### DIFF
--- a/src/ingestor/writer/proposal.ts
+++ b/src/ingestor/writer/proposal.ts
@@ -56,18 +56,25 @@ export async function verify(body): Promise<any> {
     if (msg.payload.type !== space.voting.type) return Promise.reject('invalid voting type');
   }
 
-  try {
-    const validationName = space.validation?.name || 'basic';
-    const validationParams = space.validation?.params || {};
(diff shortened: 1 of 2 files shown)


src/ingestor/writer/proposal.ts
```diff
@@ -56,18 +56,25 @@ export async function verify(body): Promise<any> {
...
-  } catch (e) {
-    return Promise.reject('failed to check validation');
+  const onlyAuthors = space.filters?.onlyMembers;
+  const authors = (space.members || []).map(address => address.toLowerCase());
+  const isAuthor = authors.includes(body.address.toLowerCase());
+  if (onlyAuthors && !isAuthor) return Promise.reject('only space authors can propose');
+
+  if (!isAuthor) {
+    try {
+      const validationName = space.validation?.name || 'basic';
```
ChaituVR, line 66: we don't need basic validation anymore no?

ChaituVR (approved):

PR #493, opened 2022-10-25
Allow alias based msgs for vote and proposal actions

bonustrack:
Change typed data ingestor to support vote and proposal action with alias

### DIFF
--- a/src/ingestor/typedData/index.ts
+++ b/src/ingestor/typedData/index.ts
@@ -40,17 +40,23 @@ export default async function ingestor(body) {
   let type = hashTypes[hash];
 
   let network = '1';
+  let aliased = false;
   if (!['settings', 'alias', 'profile'].includes(type)) {
     if (!message.space) return Promise.reject('unknown space');
     const space = await getSpace(message.space);
     if (!space) return Promise.reject('unknown space');
     network = space.network;
+    if (space.voting?.aliased) aliased = true;
   }
 
   // Check if signing address is an alias
+  const aliasTypes = ['follow', 'unfollow', 'subscribe', 'unsubscribe', 'profile'];
+  const aliasOptionTypes = ['vote', 'vote-array', 'vote-string', 'proposal', 'delete-proposal'];
   if (body.address !== message.from) {
-    if (!['follow', 'unfollow', 'subscribe', 'unsubscribe', 'profile'].includes(type))
+    if (!aliasTypes.includes(type) && !aliasOptionTypes.includes(type))
       return Promise.reject('wrong from');
+
+    if (aliasOptionTypes.includes(type) && !aliased) return Promise.reject('alias not enabled');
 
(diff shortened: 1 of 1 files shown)


src/ingestor/typedData/index.ts
```diff
@@ -40,18 +40,24 @@ export default async function ingestor(body) {
...
     if (!message.space) return Promise.reject('unknown space');
     const space = await getSpace(message.space);
     if (!space) return Promise.reject('unknown space');
     network = space.network;
+    if (space.voting?.aliased) aliased = true;
   }
 
   // Check if signing address is an alias
+  const aliasTypes = ['follow', 'unfollow', 'subscribe', 'unsubscribe', 'profile'];
+  const aliasOptionTypes = ['vote', 'vote-array', 'vote-string', 'proposal', 'delete-proposal'];
```
zzuziak, line 54: Hey @bonustrack, asking out of curiosity - why do you differentiate here between `aliasTypes` and `aliasOptionTypes`?

bonustrack: `aliasTypes` define the actions that can do with alias, `aliasOptionTypes` define the actions that can use alias only if we enable the "aliased" space settings.

src/ingestor/typedData/index.ts
```diff
@@ -103,7 +109,7 @@ export default async function ingestor(body) {
   }
 
   let legacyBody = {
-    address: body.address,
+    address: message.from,
```
ChaituVR, line 112: It won't be a problem if i pass someone's address into `from`? I have to run few tests

src/ingestor/typedData/index.ts
ChaituVR: At first I was confused too, maybe better to add a comment in code

zzuziak: How about using `aliasedTypes` and `aliasableTypes` to differentiate between the ones that are always aliased and can be aliased optionally? :thinking:

bonustrack: @zzuziak The actions "follow", "unfollow" etc can also be cast with the original address so technically both follow or vote actions can be called as aliasable, that's why I've added "option" in aliasOption but there must be better wording.

src/ingestor/typedData/index.ts
bonustrack: Run your test

src/ingestor/typedData/index.ts
zzuziak: Got it. I think it can be done in the var name without a comment. Something with `toggle`, `aliasToggableTypes`? Or simply `optional` instead of `option`

src/ingestor/typedData/index.ts
ChaituVR: Sorry for delay, this is working perfectly as far as i tested 👍

ChaituVR (approved):
Tested ACK

PR #522, opened 2022-12-01
Move to @snapshot-labs linter config

bonustrack:


package.json
```diff
@@ -8,6 +8,10 @@
     "dev": "nodemon src/index.ts",
     "start": "ts-node src/index.ts"
   },
+  "eslintConfig": {
+    "extends": "@snapshot-labs"
+  },
+  "prettier": "@snapshot-labs/prettier-config",
```
ChaituVR, line 14: @bonustrack When i run 
```
yarn prettier --write .
```
I get this error:
image

You see them too? should we add this command to package.json similar to `lint` ?

bonustrack: @ChaituVR prettier is setup with eslint, we shouldn't use prettier directly

ChaituVR (approved):
utACK

PR #546, opened 2023-02-09
Add new param plugins_contains

ChaituVR:


bonustrack (approved):
utACK

zzuziak:
Can we also add that query to `Spaces` or is it only possible at a proposal level?

Ideally it would be great to retrieve a list of spaces using SafeSnap, POAP, etc.

PR #551, opened 2023-02-21
Integrate with keycard

ChaituVR:
If you want to pass headers into the GraphQL editor, with this PR we can see a Request headers tab beside Query variables like in the below screenshot (make sure to hard reload if you don't see it)
image

Admin tasks:
- [ ] Update `KEYCARD_SECRET`

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,5 +2,5 @@ PORT=3000
 NETWORK=testnet
 DATABASE_URL=mysql://...
 SEQUENCER_URL=https://testnet.seq.snapshot.org
-KEYCARD_URL=
+KEYCARD_URL=https://keycard.snapshot.org
 KEYCARD_SECRET=
--- a/.env.example
+++ b/.env.example
@@ -3,4 +3,5 @@ NETWORK=testnet
 DATABASE_URL=mysql://...
 DEFAULT_NETWORK=1
 SEQUENCER_URL=https://testnet.seq.snapshot.org
+KEYCARD_URL=
 KEYCARD_SECRET=
--- a/.env.example
+++ b/.env.example
@@ -3,3 +3,4 @@ NETWORK=testnet
 DATABASE_URL=mysql://...
 DEFAULT_NETWORK=1
 SEQUENCER_URL=https://testnet.seq.snapshot.org
+KEYCARD_SECRET=
--- a/src/helpers/keycard.ts
+++ b/src/helpers/keycard.ts
@@ -1,0 +1,9 @@
+import { Keycard } from '@snapshot-labs/keycard';
+
+const keycard = new Keycard({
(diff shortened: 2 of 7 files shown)


src/helpers/keycard.ts
```diff
@@ -0,0 +1,21 @@
+import { Keycard } from '@snapshot-labs/keycard';
+import { sendError } from './utils';
+
+const keycard = new Keycard({
+  app: 'snapshot-hub',
+  secret: process.env.KEYCARD_SECRET || '',
+  URL: process.env.KEYCARD_URL || 'http://localhost:3007'
```
ChaituVR, line 7:
```suggestion
  URL: process.env.KEYCARD_URL || 'https://keycard.snapshot.org'
```

.env.example
```diff
@@ -2,3 +2,5 @@ PORT=3000
 NETWORK=testnet
 DATABASE_URL=mysql://...
 SEQUENCER_URL=https://testnet.seq.snapshot.org
+KEYCARD_URL=
```
ChaituVR, line 5:
```suggestion
KEYCARD_URL=https://keycard.snapshot.org
```

Todmy (changes requested):

src/helpers/keycard.ts
```diff
@@ -0,0 +1,21 @@
...
+
+const keycard = new Keycard({
+  app: 'snapshot-hub',
+  secret: process.env.KEYCARD_SECRET || '',
+  URL: process.env.KEYCARD_URL || 'https://keycard.snapshot.org'
+});
+
+const verifyKeyCard = async (req, res, next) => {
+  const key = req.headers['x-api-key'] || req.query.apiKey;
+  if (key && keycard.configured) {
```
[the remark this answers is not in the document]
ChaituVR: I don't understand what is "key is not settled" 🙈 can explain?

This line is used to check if a request has an API key or not and check if the keycard is configured or not. so that it won't block the current flow/local env if we don't have `KEYCARD_SECRET` in .env

Todmy:
the problem with a key for the `snapshot` client exists as we discussed.

Screenshot 2023-05-23 at 20 01 14

by adding a key to our FE we will make it public and contradicts the main idea of keys because anyone can steal it. 
1) as a hotfix we can use our own custom key. And will see whether someone steal this key.
2) also we can pay attention to the user agent
3) by combining 1 idea with Cloudflare we also can achieve some positive results.

In any way, it should be discussed.

ChaituVR:
API key in UI is not in scope, we can expect that we allow requests from UI with the current rate limit and no rate limit for an integrator with key

Todmy (approved):
tACK

PR #556, opened 2023-03-05
Remove DEFAULT_NETWORK env variable

ChaituVR:
We moved this env to snapshot-sequencer, so not used anywhere on the hub


bonustrack (approved):
utACK

PR #568, opened 2023-03-29
Fix default validation if minScore exist in filters

ChaituVR:
There are still a few spaces that use `space.filters.minScore` but have empty  validations, which means it is a basic validation and uses space strategies

example space:
https://hub.snapshot.org/api/spaces/gnosis.eth 

Reported by:
https://discord.com/channels/707079246388133940/1090583087156572231/1090583087156572231


bonustrack (approved):
utACK

PR #571, opened 2023-04-14
feat: add VoteWhere choice,choice_in

j0hnfl0w:
Issue https://github.com/snapshot-labs/snapshot/issues/3677
Connected https://github.com/snapshot-labs/snapshot/pull/3741#issuecomment-1507916793

- Adds choice,choice_in to VoteWhere input
- Add choice to fields in operations/votes.ts
- Fixed casing and options for orderBy, orderDirection

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -100,24 +100,29 @@ export function buildWhereQuery(fields, alias, where) {
 
     const fieldIn = where[`${field}_in`] || [];
     if (fieldIn.length > 0) {
-      query += `AND ${alias}.${field} IN (?) `;
-      params.push(fieldIn);
+      // Handle choice_in json keys search
+      if (alias === 'v' && field === 'choice') {
+        params.push(fieldIn);
+        const keys: string = fieldIn.reduce((acc: string, val: string, i: number) => {
+          const isLast = i === fieldIn.length - 1;
+
+          const param = `'$."${val}"'`;
+          params.push(param);
+
+          acc += `?${isLast ? '' : ', '}`;
+          return acc;
+        }, '');
+        query += `AND (${alias}.${field} IN (?) OR JSON_CONTAINS_PATH(${alias}.${field}, 'all', ${keys})) `;
+      } else {
+        query += `AND ${alias}.${field} IN (?) `;
+        params.push(fieldIn);
+      }
     }
 
     const fieldNotIn = where[`${field}_not_in`] || [];
     if (fieldNotIn.length > 0) {
       query += `AND ${alias}.${field} NOT IN (?) `;
       params.push(fieldNotIn);
-    }
-
-    const fieldInJsonKeys = where[`${field}_in_json_keys`] || [];
-    if (fieldInJsonKeys.length > 0) {
-      const keys: string = fieldInJsonKeys.reduce((acc: string, val: string, i: number) => {
-        const isLast = i === fieldInJsonKeys.length - 1;
-        acc += `'$."${val}"'${isLast ? '' : ', '}`;
-        return acc;
-      }, '');
-      query += `AND JSON_CONTAINS_PATH(choice, 'all', ${keys}) `;
     }
 
     if (type === 'number') {
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -100,29 +100,29 @@ export function buildWhereQuery(fields, alias, where) {
 
     const fieldIn = where[`${field}_in`] || [];
     if (fieldIn.length > 0) {
-      // Handle choice_in json keys search
-      if (alias === 'v' && field === 'choice') {
-        params.push(fieldIn);
-        const keys: string = fieldIn.reduce((acc: string, val: string, i: number) => {
-          const isLast = i === fieldIn.length - 1;
-
-          const param = `'$."${val}"'`;
-          params.push(param);
-
-          acc += `?${isLast ? '' : ', '}`;
-          return acc;
-        }, '');
-        query += `AND (${alias}.${field} IN (?) OR JSON_CONTAINS_PATH(${alias}.${field}, 'all', ${keys})) `;
-      } else {
-        query += `AND ${alias}.${field} IN (?) `;
-        params.push(fieldIn);
-      }
+      query += `AND ${alias}.${field} IN (?) `;
+      params.push(fieldIn);
     }
 
     const fieldNotIn = where[`${field}_not_in`] || [];
     if (fieldNotIn.length > 0) {
       query += `AND ${alias}.${field} NOT IN (?) `;
       params.push(fieldNotIn);
+    }
+
+    const fieldChoiceMatch = where[`${field}_match`] || [];
+    if (alias === 'v' && fieldChoiceMatch.length > 0) {
+      params.push(fieldChoiceMatch);
+      const keys: string = fieldChoiceMatch.reduce((acc: string, val: string, i: number) => {
+        const isLast = i === fieldChoiceMatch.length - 1;
+
+        const param = `'$."${val}"'`;
+        params.push(param);
+
+        acc += `?${isLast ? '' : ', '}`;
+        return acc;
+      }, '');
+      query += `AND (${alias}.${field} IN (?) OR JSON_CONTAINS_PATH(${alias}.${field}, 'all', ${keys})) `;
     }
 
     if (type === 'number') {
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -110,6 +110,16 @@ export function buildWhereQuery(fields, alias, where) {
       params.push(fieldNotIn);
     }
 
+    const fieldInJsonKeys = where[`${field}_in_json_keys`] || [];
+    if (fieldInJsonKeys.length > 0) {
+      const keys: string = fieldInJsonKeys.reduce((acc: string, val: string, i: number) => {
+        const isLast = i === fieldInJsonKeys.length - 1;
+        acc += `'$."${val}"'${isLast ? '' : ', '}`;
+        return acc;
+      }, '');
+      query += `AND JSON_CONTAINS_PATH(choice, 'all', ${keys}) `;
+    }
+
     if (type === 'number') {
       const fieldGt = where[`${field}_gt`];
       const fieldGte = where[`${field}_gte`];
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -111,7 +111,7 @@ export function buildWhereQuery(fields, alias, where) {
     }
 
     const fieldChoiceMatch = where[`${field}_match`] || [];
-    if (alias === 'v' && fieldChoiceMatch.length > 0) {
+    if (fieldChoiceMatch.length > 0 && alias === 'v') {
       params.push(fieldChoiceMatch);
       const keys: string = fieldChoiceMatch.reduce((acc: string, val: string, i: number) => {
         const isLast = i === fieldChoiceMatch.length - 1;
--- a/src/graphql/operations/votes.ts
+++ b/src/graphql/operations/votes.ts
@@ -29,10 +29,10 @@ async function query(parent, args, context?, info?) {
 
   let orderBy = args.orderBy || 'created';
   let orderDirection = args.orderDirection || 'desc';
-  if (!['created', 'vp'].includes(orderBy)) orderBy = 'created';
+  if (!['created', 'vp', 'choice'].includes(orderBy)) orderBy = 'created';
   orderBy = `v.${orderBy}`;
   orderDirection = orderDirection.toUpperCase();
-  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+  if (!['asc', 'desc'].includes(orderDirection)) orderDirection = 'desc';
 
   let votes: any[] = [];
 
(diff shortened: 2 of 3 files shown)


samuveth (approved):

samuveth (approved):
utACK

src/graphql/operations/votes.ts
```diff
@@ -20,18 +20,19 @@ async function query(parent, args, context?, info?) {
...
   const params: any[] = whereQuery.params;
 
   let orderBy = args.orderBy || 'created';
   let orderDirection = args.orderDirection || 'desc';
-  if (!['created', 'vp'].includes(orderBy)) orderBy = 'created';
+  if (!['created', 'vp', 'choice'].includes(orderBy)) orderBy = 'created';
   orderBy = `v.${orderBy}`;
   orderDirection = orderDirection.toUpperCase();
-  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+  if (!['asc', 'desc'].includes(orderDirection)) orderDirection = 'desc';
```
ChaituVR, line 35: 🤔 Why do we need this? we are doing this in the line above
```js
orderDirection = orderDirection.toUpperCase();
```

ChaituVR (changes requested):

src/graphql/schema.gql
```diff
@@ -210,6 +210,8 @@ input VoteWhere {
   vp_lte: Float
   vp_state: String
   vp_state_in: [String]
+  choice: Any
+  choice_in: [Any]
```
ChaituVR, line 214: I think the input choice can be a string, I am not sure if it will work with an object in case it is a weighted voting, but I will test it soon
```suggestion
  choice: String
  choice_in: [String]
```

src/graphql/operations/votes.ts
```diff
@@ -20,18 +20,19 @@ async function query(parent, args, context?, info?) {
     app: 'string',
     created: 'number',
     vp: 'number',
-    vp_state: 'string'
+    vp_state: 'string',
+    choice: 'any'
```
ChaituVR, line 24:
```suggestion
    choice: 'string'
```

src/graphql/schema.gql
ChaituVR: Hmm looks like doesn't work for weighted voting choices, Have you tried? should pass an object or a string?

j0hnfl0w: I tested locally only with strings

j0hnfl0w: Can you share some proposals with weighted voting choices?

ChaituVR: Example https://snapshot.org/#/aavegotchi.eth/proposal/0xade68c6cf0285a55e4cf914266761b9a505389ef4686e7972ad5426ac49cd238

j0hnfl0w: Thanks!
So we need filter like this
Types will be:  choice: string, choice_in: string[]
```
Object.keys(choice).filter(value => filter.choice_in.includes(value))
```

But we need to do it with sql [here](https://github.com/snapshot-labs/snapshot-hub/blob/master/src/graphql/helpers.ts#L101)

ChaituVR: I am bit confused but `buildWhereQuery` is used in many places we should make sure others are not affected 😅

ChaituVR: @bonustrack Any advice?

samuveth: Any chance we can move forward with this without @bonustrack?

ChaituVR: So as far as I remember Fabien was against the JSON search functions so want to confirm with him. Do you see any alternative ways @j0hnfl0w?

j0hnfl0w: Alternatives:
1. Search all votes client-side, seems impossible to preload 60k votes
2. Search by choice only for votes with simple choices (string/number)

ChaituVR: Maybe is it possible to convert column to string instead of JSON @bonustrack

j0hnfl0w: Even if we convert all these json to string how we are gonna search by json keys?

ChaituVR: Ahh yes, hiding my comment, then only JSON search should work. maybe you can continue with it. as Fabien not replying 😄

j0hnfl0w:
@ChaituVR so here is a solution with JSON_CONTAINS_PATH, i tested it locally and it works.
How to test in on testnet db?

j0hnfl0w:
@ChaituVR ping

src/graphql/schema.gql
```diff
@@ -210,6 +210,9 @@ input VoteWhere {
   vp_lte: Float
   vp_state: String
   vp_state_in: [String]
+  choice: String
+  choice_in: [String]
+  choice_in_json_keys: [String]
```
ChaituVR, line 215: We are not using this anywhere right?
```suggestion
```

src/graphql/helpers.ts
```diff
@@ -110,6 +110,16 @@ export function buildWhereQuery(fields, alias, where) {
...
     }
 
+    const fieldInJsonKeys = where[`${field}_in_json_keys`] || [];
+    if (fieldInJsonKeys.length > 0) {
+      const keys: string = fieldInJsonKeys.reduce((acc: string, val: string, i: number) => {
+        const isLast = i === fieldInJsonKeys.length - 1;
+        acc += `'$."${val}"'${isLast ? '' : ', '}`;
+        return acc;
+      }, '');
+      query += `AND JSON_CONTAINS_PATH(choice, 'all', ${keys}) `;
```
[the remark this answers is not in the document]
j0hnfl0w: We will apply choice_in_json_keys filter only for weighted voting for others we will use choice_in filter.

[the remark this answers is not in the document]
ChaituVR: Should be the same for weighted, Quadratic voting too - they are objects too

But approval and approval voting use arrays, for example `[4, 3, 2, 1]`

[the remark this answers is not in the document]
j0hnfl0w: So we decide on frontend what filter to use choice_in or choice_in_json_keys(weighted, quadratic)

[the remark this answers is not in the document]
ChaituVR: Can test these in your local?
Also
```
      query += `AND JSON_CONTAINS_PATH(choice, 'all', ${keys}) `;
 ```
 
 This line is not good 🙈 someone may write some SQL code to delete everything in DB
should pass it into the `params` variable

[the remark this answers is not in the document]
Comment (j0hnfl0w) [reply src/graphql/helpers.ts:120]: yep i tested it locally.
to do sql injection here we need to pass it with keys like ["delete", "from spaces"] ?
yeah will move it to params to sanitize them

File: src/graphql/schema.gql (hunk 4 above)
Comment (ChaituVR) [inline src/graphql/schema.gql:215]: Instead of a separate variable for objects and seperate variable for strings, can we just check both with one variable?

Comment (ChaituVR) [reply src/graphql/schema.gql:215]: I mean maybe we can handle it outside buildWhereQuery, once we pass a string, it should check both as string and JSON

File: src/graphql/helpers.ts (hunk 5 above)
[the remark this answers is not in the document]
Comment (j0hnfl0w) [reply src/graphql/helpers.ts:120]: So will it work like this?
We need AND JSON_CONTAINS_PATH(choice, 'all', ?, ?, ?)
And push to params ['$."1"', '$."2"']
```js
const fieldInJsonKeys = where[`${field}_in_json_keys`] || [];
if (fieldInJsonKeys.length > 0) {
  const keys: string = fieldInJsonKeys.reduce((acc: string, val: string, i: number) => {
      const isLast = i === fieldInJsonKeys.length - 1;
    
      const param = `'$."${val}"'`;
      params.push(param);
    
      acc += `? ${isLast ? '' : ','}`;
      return acc;
  }, '');
  query += `AND JSON_CONTAINS_PATH(choice, 'all', ${keys}) `;
}
```

src/graphql/helpers.ts
```diff
@@ -100,8 +100,23 @@ export function buildWhereQuery(fields, alias, where) {
 
     const fieldIn = where[`${field}_in`] || [];
     if (fieldIn.length > 0) {
-      query += `AND ${alias}.${field} IN (?) `;
-      params.push(fieldIn);
+      // Handle choice_in json keys search
+      if (alias === 'v' && field === 'choice') {
+        params.push(fieldIn);
```
ChaituVR, line 105: If someone requests with `choice: '1'` it works for single choice voting only but if `choice_in: ['1']` it works for all voting types. 

Both should be similar, they are just different inputs, outputs should be same

ChaituVR: Otherwise looks good 👍

samuveth:
Please approve if all is good

ChaituVR:
> Please approve if all is good

Please refer to my comments :) https://github.com/snapshot-labs/snapshot-hub/pull/571#discussion_r1181452186

src/graphql/helpers.ts
j0hnfl0w: choice is exact match where query, you mean when user search with choice: '1' we convert it to choice_in: ['1']?
Maybe we can remove this exact search query due to complex search of json choices?

ChaituVR: I am more confused now 😆 

`choice_in` should be the same as `choice`, choice_in just that we pass more inputs

1. If we need an exact match, then `choice_in` should also do an exact match too, but these won't be useful for UI right? works only for single choice

2. Maybe we can have a new option `choice_matches` like before but which can search the JSON for both exact match and JSON choice match and leave `choice` and `choice_in` for an exact match

j0hnfl0w: So like it was from the start with choice_in_json_keys? 😃

ChaituVR: hehe yea but `choice_in_json_keys` was just for choice inside JSON keys, we should also check for an exact match, the name also confuses 🙈  maybe `choice_match` or something like that should be good i think

j0hnfl0w: Maybe name it ```choice_contains_path``` after JSON_CONTAINS_PATH ?

ChaituVR: that name mean it won't check for an exact match 🙈

j0hnfl0w: Now we have:
choice - search for exact match (does not work for json choice)
choice_in - search for keys in json OR match in array
How do want to change that?
choice - search for exact match (does not work for json choice)
choice_in - search for match in array (does not work for json choice)
choice_matches - search for keys in json OR match in array

src/graphql/schema.gql
```diff
@@ -210,6 +210,9 @@ input VoteWhere {
   vp_lte: Float
   vp_state: String
   vp_state_in: [String]
+  choice: Any
+  choice_in: [Any]
+  choice_match: [String]
```
ChaituVR, line 215: `choice_match` can still search exact match right? if so we don't need `choice` and `choice_in`

ChaituVR: if not, we should match the exact match

samuveth:
@ChaituVR @j0hnfl0w what's the status here?

j0hnfl0w:
Added last improvements, need to review and test it @ChaituVR

src/graphql/operations/votes.ts
```diff
@@ -20,15 +20,16 @@ async function query(parent, args, context?, info?) {
...
+    choice: 'string'
   };
   const whereQuery = buildWhereQuery(fields, 'v', where);
   const queryStr = whereQuery.query;
   const params: any[] = whereQuery.params;
 
   let orderBy = args.orderBy || 'created';
   let orderDirection = args.orderDirection || 'desc';
-  if (!['created', 'vp'].includes(orderBy)) orderBy = 'created';
+  if (!['created', 'vp', 'choice'].includes(orderBy)) orderBy = 'created';
```
bonustrack, line 32: Why do we need order by choice?

j0hnfl0w: > Additional filter options to consider include sorting by ascending voting power, descending voting power, and by choice

https://github.com/snapshot-labs/snapshot/issues/3677#issue-1651211014

src/graphql/helpers.ts
```diff
@@ -110,6 +110,21 @@ export function buildWhereQuery(fields, alias, where) {
...
+      const keys: string = fieldChoiceMatch.reduce((acc: string, val: string, i: number) => {
+        const isLast = i === fieldChoiceMatch.length - 1;
+
+        const param = `'$."${val}"'`;
+        params.push(param);
+
+        acc += `?${isLast ? '' : ', '}`;
+        return acc;
+      }, '');
+      query += `AND (${alias}.${field} IN (?) OR JSON_CONTAINS_PATH(${alias}.${field}, 'all', ${keys})) `;
```
bonustrack, line 125: Wouldn't this be exposed to SQL injection? I don't see how value get parsed on the JSON_CONTAINS_PATH query, someone may bring something like this as value `")); TRUNCATE proposals;"` and delete proposals table isn't?

samuveth: Some input from GPT-4
```
The code does risk SQL injection through the JSON_CONTAINS_PATH part. Although it uses parameterized queries, the keys variable is directly interpolated into the SQL string. This can lead to SQL injection if these values come from an untrusted source like user input. It's safer to use parameter binding for JSON_CONTAINS_PATH, if possible with your SQL library.
```

Todmy (commented):
should we close it?

PR #576, opened 2023-04-24
feat: Order by rank

ChaituVR:
Related to https://github.com/snapshot-labs/snapshot/issues/1437

- [x] Order by rank
- [x] Ignores private space if sorted by rank
- [x] Search with the id string
- [ ] Search with  `category` and `network` 
- [ ] New Query to get metrics like total spaces, total networks, total strategies, and total plugins

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "snapshot-spaces"]
+	path = snapshot-spaces
+	url = https://github.com/snapshot-labs/snapshot-spaces
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -2,16 +2,34 @@ import snapshot from '@snapshot-labs/snapshot.js';
 import { uniq } from 'lodash';
 import db from './mysql';
 import log from './log';
+import verifiedSpaces from '../../snapshot-spaces/spaces/verified.json';
 
 export let spaces = {};
+export const exploreEndpointData = {};
 export const spacesMetadata = {};
 export const spaceProposals = {};
 export const spaceVotes = {};
 export const spaceFollowers = {};
 
+function getRank(id: string, verified: number): number {
+  let ranking =
+    (spaceVotes[id]?.count || 0) / 50 +
+    (spaceVotes[id]?.count_7d || 0) +
+    (spaceProposals[id]?.count_7d || 0) * 50 +
+    (spaceFollowers[id]?.count_7d || 0);
+
+  if (verified) {
+    ranking = ranking * 5;
+    ranking += 100;
+  }
+
+  return ranking;
+}
+
 function mapSpaces() {
   Object.entries(spaces).forEach(([id, space]: any) => {
-    spacesMetadata[id] = {
+    /* This will be deprecated soon */
+    exploreEndpointData[id] = {
       name: space.name,
       private: space.private || undefined,
       terms: space.terms || undefined,
@@ -26,6 +44,14 @@ function mapSpaces() {
       votes_7d: (spaceVotes[id] && spaceVotes[id].count_7d) || undefined,
       followers: (spaceFollowers[id] && spaceFollowers[id].count) || undefined,
       followers_7d: (spaceFollowers[id] && spaceFollowers[id].count_7d) || undefined
+    };
+    /* This will be deprecated soon */
+
+    spacesMetadata[id] = {
+      id,
+      verified: verifiedSpaces[id] || 0,
+      rank: getRank(id, verifiedSpaces[id]),
+      private: space.private || undefined
     };
   });
 }
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,5 +1,5 @@
 import express from 'express';
-import { spaces, spacesMetadata } from './helpers/spaces';
+import { spaces, exploreEndpointData } from './helpers/spaces';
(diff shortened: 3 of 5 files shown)


ChaituVR (commented):
To do as mentioned in https://github.com/snapshot-labs/snapshot/issues/3863#issuecomment-1534226992
- [x] Return the total count of the matching query (if I search with a network, we should also return the total number of spaces matching that network)
- [x] `skip` doesn't look like working - some bug
- [x] add a new input `search` that matches both name or id, remove `id_contains`

ChaituVR:
Also added
- [x] Spaces per category
- [x] Spaces per network
- [x] Spaces per strategy
- [x] Spaces per Plugin
- [x] Remove explore endpoint

ChaituVR:
- Added new query `activeSpaces` this won't return private spaces, so strictly for home page :sweat_smile: 
- Also doesn't require `private` in `where` 
- Also removed `id` and `id_in` in `where` as we won't use them
- Make sure sidebar use old spaces query itself
- Removed `metrics` query
- `activeSpaces` return `categories` count with matching filters (also doesn't count private space categories)
- Queries networks, strategies, etc still return count per space, but also include private space's networks, strategies etc

bonustrack:
I feel the name `activeSpaces` is confusing, spaces returned are not necessary active, and we don't really return spaces, can we call it `ranking` instead?

src/graphql/operations/activeSpaces.ts
```diff
@@ -0,0 +1,16 @@
+import { checkLimits, fetchActiveSpaces, handleRelatedSpaces, PublicError } from '../helpers';
+import log from '../../helpers/log';
+
+export default async function (_parent, args, _context, info) {
+  checkLimits(args, 'activeSpaces');
+  try {
+    const { spaces, metrics } = await fetchActiveSpaces(args);
```
bonustrack, line 7: Can we move the function `fetchActiveSpaces` on this file? It's only being used here, it's better if we keep the logic in same place and use helpers for functions that are used on different files.

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "snapshot-spaces"]
+	path = snapshot-spaces
+	url = https://github.com/snapshot-labs/snapshot-spaces
```
bonustrack, line 3: Would be better if we can move verified.json to sidekick so we don't need to redeploy the hub everytime there is a change there.

samuveth: Data from sidekick isn't ready yet, but will be soon hopefully

samuveth: Can we remove this submodule now? @ChaituVR

ChaituVR:
- Removed  snapshot-spaces submodule
- Using sidekick to get all moderation data
- Now we return flagged property in spaces and proposals
- Now we can filter proposals query with space_verified:true in where,  so it can return proposals for only for verified spaces, so don't need to pass spaces_in in timeline, for "joined spaces"  timeline, while fetching follows just filter for verified spaces and pass to proposals

samuveth:
Great! Can we also have a way to filter out flagged proposals using `where`?

ChaituVR:
So a lot of things changes from the plan, Here is the summary of the changes:

1. Removed explore endpoint
2. Added a new query for the home page - we call it `ranking` 
3. Ranking query never consider private spaces, but filters spaces with the network, a search string (id or name), and category - As per needs of the home page
4. Ranking query also sorts spaces with ranks, ranks are calculated with popularity (calculating popularity is same as before - which was done on UI)
5. We now request all moderation data from the sidekick server and mark spaces/proposals as verified/flagged
6. Added a new filter to the Proposals query, to request proposals from verified spaces only (required for the timeline page)
7. Every space object will now return these fields `verified`, `flagged`, and `rank` and also counts like:
  ```
        activeProposals
        proposalsCount
        proposalsCount7d
        followersCount
        followersCount7d
        votesCount
        votesCount7d
 ```

To test these changes, you can use this query
```Graphql
{
  ranking(first: 20, skip: 0, where:{network: "137", category: "service"})
  {
    metrics {
      total
      categories
    },
    items {
      id
      name
      private
      verified
      flagged
      categories
      rank
      votesCount
      votesCount7d
      followersCount
      followersCount7d
    }
  }
}
```

src/helpers/spaces.ts
```diff
@@ -2,32 +2,68 @@ import snapshot from '@snapshot-labs/snapshot.js';
...
+        followersCount7d: spaceFollowers[id]?.count_7d || 0,
+        votesCount: spaceVotes[id]?.count || 0,
+        votesCount7d: spaceVotes[id]?.count_7d || 0
+      }
     };
   });
+
+  rankedSpaces = Object.values(spacesMetadata)
+    .sort((a: any, b: any) => b.popularity - a.popularity)
+    .filter((space: any) => !space.private);
```
Todmy, line 62: Should we hide suspicious spaces as well?

Todmy (changes requested):

src/helpers/spaces.ts
ChaituVR: Yea there is no confirmation from @bonustrack Maybe we can hide them in future, for this PR we don't need to hide flagged spaces

Todmy (approved):
tACK

PR #585, opened 2023-05-17
Hide flagged spaces and check networks and strategies while calculating popularity

ChaituVR:
Changes:
- Hiding flagged spaces
- If space doesn't use any testnet network then multiply popularity by 3
- If space doesn't use any ticket strategy then multiple popularity by 3

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -1,4 +1,5 @@
 import snapshot from '@snapshot-labs/snapshot.js';
+import networks from '@snapshot-labs/snapshot.js/src/networks.json';
 import { uniq } from 'lodash';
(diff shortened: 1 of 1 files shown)


src/helpers/spaces.ts
```diff
@@ -12,17 +13,28 @@ const spaceProposals = {};
 const spaceVotes = {};
 const spaceFollowers = {};
 
-function getPopularity(id: string, verified: boolean): number {
+const testnetNetworks = Object.values(networks)
```
samuveth, line 16: Would call it `testnets` or `testNetworks`

Todmy (approved):
utACK: not a lot of changes, validated and tried locally

PR #593, opened 2023-05-27
fix: update graphql-tools

bonustrack:
Currently running the hub in local throw an error
image
This PR fix this, also remove constrain on node version. Need to be tested in DO before merge


ChaituVR (changes requested):
Now i get this warning 

```text
This package has been deprecated and now it only exports makeExecutableSchema.
And it will no longer receive updates.
We recommend you to migrate to scoped packages such as @graphql-tools/schema, @graphql-tools/utils and etc.
Check out https://www.graphql-tools.com to learn what package you should use instead!
```

PR #618, opened 2023-07-27
feat: setup sentry

wa0x6e:
- Add sentry
- Create sourcemaps on build
- Upload the sourcemaps and create a Sentry release when deploying on main/master branch


Todmy (approved):
utACK

PR #620, opened 2023-07-27
chore: add build workflow

wa0x6e:

The current production app is run directly with `ts-node` on typescript file, instead of with `node` on compiled js files.

This result in 10x more memory consumption.

My guess is that that solution was retained because the build workflow (`tsc`) does not copy the `gql` file to the final build directory.

## 💊 Fixes / Solution

- Run production app with `node` instead of `ts-node`

## 🚧 Changes

- Build files and run the production app with `node` command
- Add a post-build task to copy the `gql` file to the build directory
 
## 🛠️ Tests

- Run `yarn build`
- Run `yarn start`
- It should start the app


ChaituVR (approved):
tAck

PR #624, opened 2023-07-29
Query to filter flagged proposals

ChaituVR:
How to test:
- To return all proposals:
  ```
  {
    proposals(first:1000){
      id
      space {
        id
      }
      flagged
    }
  }
  ```

- To return only flagged proposals:
  ```
  {
    proposals(first:1000, where:{flagged:true}){
      id
      space {
        id
      }
      flagged
    }
  }
  ```
- To return only nonflagged proposals:
  ```
  {
    proposals(first:1000, where:{flagged:false}){
      id
      space {
        id
      }
      flagged
    }
  }
  ```

### DIFF
--- a/src/graphql/operations/proposals.ts
+++ b/src/graphql/operations/proposals.ts
@@ -66,7 +66,7 @@ export default async function (parent, args) {
   }
 
   if (where.flagged === true) {
-    searchSql += ' AND p.id in (?)';
+    searchSql += ' AND p.id IN (?)';
     params.push(flaggedProposals);
   }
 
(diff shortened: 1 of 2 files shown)


src/graphql/operations/proposals.ts
```diff
@@ -65,6 +65,16 @@ export default async function (parent, args) {
     params.push(verifiedSpaces);
   }
 
+  if (where.flagged === true) {
+    searchSql += ' AND p.id in (?)';
```
bonustrack, line 69: Sql arguments should be in uppercase
```suggestion
    searchSql += ' AND p.id IN (?)';
```

src/graphql/operations/proposals.ts
```diff
@@ -65,6 +65,16 @@ export default async function (parent, args) {
...
+    params.push(flaggedProposals);
+  }
+
+  if (where.flagged === false) {
+    searchSql += ' AND p.id not in (?)';
```
bonustrack, line 74:
```suggestion
    searchSql += ' AND p.id NOT IN (?)';
```

samuveth (approved):
utACK

PR #625, opened 2023-08-01
chore: improve CI

wa0x6e:
- Migrate workflows to use snapshot-labs/actions
- Rename existing workflows to `verb-description` convention


Todmy (approved):
utACK

PR #632, opened 2023-08-03
chore: allow a certain number of fails before reporting the error to Sentry

wa0x6e:

Retrieving data frequently from our own some services (sidekick and score) can temporary fail for various reasons (deploy progress, etc ...)

From the sentry logs, we can see that these fail at least once a day. 

As the data is already cached, and the query is for updating it periodically, the query is allowed to fail a few times without real impact, and should raise an error only when it fails consecutively.

## 💊 Fixes / Solution

All temporary network errors should be muted, unless they're raised consecutively more than N times (which means something really wrong, and not temporary). 

This will remove a lot of noises from our error reporting.

Fixes SNAPSHOT-HUB-5
Fixes SNAPSHOT-HUB-6
Fixes SNAPSHOT-HUB-8

## 🚧 Changes

- Add a counter to store the number of times the query fail
- When a query fails consecutively more than n time, report the error to senrtry

### DIFF
--- a/src/helpers/moderation.ts
+++ b/src/helpers/moderation.ts
@@ -3,6 +3,7 @@ import log from './log';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 
 const moderationURL = 'https://sh5.co/api/moderation';
+let consecutiveFailsCount = 0;
 
 export let flaggedLinks: Array<string> = [];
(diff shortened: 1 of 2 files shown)


ChaituVR (approved):
utAck

PR #634, opened 2023-08-04
Return limits and remaining count in headers

ChaituVR:
🔴 Need a new version from https://github.com/snapshot-labs/keycard.js/pull/7

Summary of changes:

- Use new values returned from keycard.js
- Return remaining count, limits and reset time in headers
  Untitled 3
- Fallback to default rate limit if the max number of requests on keycard is reached

How to test:
1. Install the new version of keycard.js with `yarn link` or wait for the new version to be released
2. Run the server
3. Go to `http://localhost:3000/graphql?apiKey=1234` (replace apiKey if required)
4. Try sending the requests and look into the networks tab in console
5. It should display the correct limits and remaining count
6. If it reaches our limit, it should fall back to the default rate limit (express-rate-limit)

### DIFF
--- a/src/helpers/keycard.ts
+++ b/src/helpers/keycard.ts
@@ -7,15 +7,18 @@ const keycard = new Keycard({
   URL: process.env.KEYCARD_URL || 'https://keycard.snapshot.org'
 });
 
-const verifyKeyCard = async (req, res, next) => {
-  const key = req.headers['x-api-key'] || req.query.apiKey;
-  if (key && keycard.configured) {
-    const { valid, rateLimited } = keycard.logReq(key);
+const checkKeycard = async (req, res, next) => {
+  const apiKey = req.headers['x-api-key'] || req.query.apiKey;
+  if (apiKey && keycard.configured) {
+    const keycardData = keycard.logReq(apiKey);
+    if (!keycardData.valid) return sendError(res, 'invalid api key', 401);
 
-    if (!valid) return sendError(res, 'invalid api key', 401);
-    if (rateLimited) return sendError(res, 'too many requests', 429);
+    res.locals.keycardData = keycardData;
+    res.set('X-Api-Key-Limit', keycardData.limit);
+    res.set('X-Api-Key-Remaining', keycardData.remaining);
+    res.set('X-Api-Key-Reset', keycardData.reset);
   }
   return next();
 };
 
-export { keycard, verifyKeyCard };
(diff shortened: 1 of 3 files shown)


src/helpers/keycard.ts
```diff
@@ -7,15 +7,18 @@ const keycard = new Keycard({
...
+  if (apiKey && keycard.configured) {
+    const keycardData = keycard.logReq(apiKey);
+    if (!keycardData.valid) return sendError(res, 'invalid api key', 401);
 
-    if (!valid) return sendError(res, 'invalid api key', 401);
-    if (rateLimited) return sendError(res, 'too many requests', 429);
+    res.locals.keycardData = keycardData;
+    res.set('X-Api-Key-Limit', keycardData.limit);
+    res.set('X-Api-Key-Remaining', keycardData.remaining);
+    res.set('X-Api-Key-Reset', keycardData.reset);
```
ChaituVR, line 19: These lint errors should be fixed once we install new version of keycard.js

src/helpers/keycard.ts
```diff
@@ -7,15 +7,18 @@ const keycard = new Keycard({
...
 
-const verifyKeyCard = async (req, res, next) => {
-  const key = req.headers['x-api-key'] || req.query.apiKey;
-  if (key && keycard.configured) {
-    const { valid, rateLimited } = keycard.logReq(key);
+const checkKeycard = async (req, res, next) => {
+  const apiKey = req.headers['x-api-key'] || req.query.apiKey;
+  if (apiKey && keycard.configured) {
+    const keycardData = keycard.logReq(apiKey);
+    if (!keycardData.valid) return sendError(res, 'invalid api key', 401);
```
wa0x6e, line 14: Isn't this always returning `true` from keycard.js ? https://github.com/snapshot-labs/keycard.js/blob/f3bcc6a5b183bf02ba536127ac579656cae25d5d/src/index.ts#L77

src/helpers/keycard.ts
wa0x6e: Let's merge and publish the new keycard.js version first, can not even `yarn dev` because of typescript error

ChaituVR: Using keycard.js from github for now. will use from npm once it is published 

But it is not working on this PR because it is private repo 😃

wa0x6e (approved):
tAck

PR #635, opened 2023-08-07
feat(env): add environment variables for configuring score api and sidekick

Todmy:


src/helpers/strategies.ts
```diff
@@ -6,7 +6,8 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
 export let strategies: any[] = [];
 export let strategiesObj: any = {};
 
-const uri = 'https://score.snapshot.org/api/strategies';
+const scoreApiEndpoint = process.env.SCORE_API_URL || 'https://score.snapshot.org';
```
wa0x6e, line 9: Shouldn't this be named `scoreApiURL` like the rest?

wa0x6e (changes requested):

src/helpers/strategies.ts
Todmy: fixed

wa0x6e (approved):
utAck

PR #650, opened 2023-08-18
chore: fix wrong context key

wa0x6e:
- Fix wrong contexts key on existing sentry capture 
- Add more contexts to error capture


src/graphql/operations/messages.ts
```diff
@@ -36,7 +36,7 @@ export default async function (parent, args) {
     return await db.queryAsync(query, params);
   } catch (e: any) {
     log.error(`[graphql] messages, ${JSON.stringify(e)}`);
-    capture(e);
+    capture(e, { contexts: { input: { args } } });
```
Todmy, line 39: I see that it has the same usage everywhere. Don't you want to simplify the interface? It could be this way `capture(e, { args })`

Todmy (changes requested):

src/graphql/operations/messages.ts
wa0x6e: we do need the `input` key, as it's the category. And `contexts` tell us what kind of context it is, it can be `tags` of other things. We're mainly using `contexts` for now.

I agree it's cumbersome that it needs 2 levels depth, but it's also the original `captureException` function signature by the sentry SDK

ChaituVR (approved):
utAck

Todmy (approved):
utACK

PR #653, opened 2023-08-24
feat: add user_space table to collect user's votes and proposals count per space

wa0x6e:
From the pitches https://www.notion.so/snapshotlabs/S-delegation-part-2-2b3186d8a33b4fd8b90c6575b6bf8000#bc7c1c9c785c46bdad69b4c8337771ca

This is a read-only part of the feature, which creates a new table to store the `proposals_count` and `votes_count` counters. The update side will be on sequencer: https://github.com/snapshot-labs/snapshot-sequencer/pull/127

This PR contains 2 changes: 

- Update to the schema.sql to include the new table
- ~~Add a script (run with `yarn ts-node scripts/generate_user_space_counters.ts`) to trigger the counters computation.~~ (moved to sequencer)


src/helpers/schema.sql
```diff
@@ -142,3 +142,14 @@ CREATE TABLE statements (
   INDEX created (created),
   INDEX updated (updated)
 );
+
+CREATE TABLE user_space_activities (
+  user_id varchar(64) NOT NULL,
+  space_id varchar(64) NOT NULL,
+  votes_count smallint unsigned NOT NULL DEFAULT '0',
+  proposals_count smallint unsigned NOT NULL DEFAULT '0',
```
ChaituVR, line 150: We need to uppercase `varchar` like above

wa0x6e:
Closing, see leaderboard instead

PR #658, opened 2023-08-28
chore: relax dependabot rule

wa0x6e:

Dependabot only updates package for snapshot.js. It should also monitor all other packages from  @snapshot-labs orgs

## 💊 Fixes / Solution

Make dependabot create PR on new packages from the same org

## 🚧 Changes

- relax dependabot rule to include all packages from @snapshot-labs

## 🛠️ Tests

- N/A


ChaituVR (approved):
utAck

PR #663, opened 2023-08-30
chore: instrument open connections count

wa0x6e:
This PR add metrics to track the number of open connections to the expressjs server


ChaituVR (approved):
utAck

PR #669, opened 2023-09-07
feat(proposal): add ability to update proposals

Todmy:
This PR requires adding new column `updated INT(11) DEFAULT NULL` in proposals table

### DIFF
--- a/src/helpers/schema.sql
+++ b/src/helpers/schema.sql
@@ -19,6 +19,7 @@ CREATE TABLE proposals (
   ipfs VARCHAR(64) NOT NULL,
   author VARCHAR(64) NOT NULL,
   created INT(11) NOT NULL,
+  updated INT(11) DEFAULT NULL,
   space VARCHAR(64) NOT NULL,
   network VARCHAR(12) NOT NULL,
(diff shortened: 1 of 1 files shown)


src/helpers/schema.sql
```diff
@@ -19,6 +19,7 @@ CREATE TABLE proposals (
   ipfs VARCHAR(64) NOT NULL,
   author VARCHAR(64) NOT NULL,
   created INT(11) NOT NULL,
+  updated INT(11) DEFAULT NULL,
```
wa0x6e, line 22: Maybe also add an `INDEX` for this field ?

Todmy: not sure. Do you think we need this index?

wa0x6e: There's an option to use on WHERE query such as updated ==, !=, >, and <, so I think it'll benefit from an index

wa0x6e (approved):
tAck

src/helpers/schema.sql
ChaituVR, line 22: `created` doesn't have DEFAULT? 🤔

ChaituVR (approved):

PR #679, opened 2023-09-21
feat(flag proposal): add ability to flag proposal

Todmy:
- Fix query builder to get flagged/not flagged proposals
- modified graphql helper
- fix SQL schema

TODO:
- [x] Create a new column with 
```sql
ALTER TABLE proposals 
ADD COLUMN flagged INT NOT NULL DEFAULT 0,
ADD INDEX flagged (flagged);
```

Notes:
This PR doesn't remove the previous implementation and is backward compatible. The previous implementation will be removed when data is migrated from the laser DB to the new column. Probably it will be done in a separate PR

### DIFF
--- a/src/graphql/operations/proposals.ts
+++ b/src/graphql/operations/proposals.ts
@@ -66,13 +66,15 @@ export default async function (parent, args) {
     params.push(verifiedSpaces);
   }
 
+  // TODO: remove part `p.id IN (?)` when flagged proposals are moved from laser DB to snapshot-sequencer DB
   if (where.flagged === true && flaggedProposals.length > 0) {
-    searchSql += ' AND p.id IN (?)';
+    searchSql += ' AND (p.id IN (?) OR p.flagged = 1)';
     params.push(flaggedProposals);
   }
 
+  // TODO: remove part `p.id NOT IN (?)` when flagged proposals are moved from laser DB to snapshot-sequencer DB
   if (where.flagged === false && flaggedProposals.length > 0) {
-    searchSql += ' AND p.id NOT IN (?)';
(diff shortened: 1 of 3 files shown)


src/graphql/operations/proposals.ts
```diff
@@ -66,13 +66,15 @@ export default async function (parent, args) {
     params.push(verifiedSpaces);
   }
 
+  // TODO: remove part `p.id IN (?)` when flagged proposals are moved from laser DB to snapshot-sequencer DB
   if (where.flagged === true && flaggedProposals.length > 0) {
```
wa0x6e, line 70: Beware of the case where this condition will not be triggered if `flaggedProposals.length` is 0.

Refactoring to remove `p.id IN (?)` has to be done before disconnecting laser database

wa0x6e (dismissed):
tAck

src/helpers/schema.sql
```diff
@@ -58,6 +59,7 @@ CREATE TABLE proposals (
   INDEX scores_state (scores_state),
   INDEX scores_updated (scores_updated),
   INDEX votes (votes)
```
wa0x6e, line 61: Missing trailing comma here

wa0x6e (changes requested):

src/graphql/operations/proposals.ts
Todmy: I know, thank you 🙏

Todmy (commented):
fix(mysql): fix error in schema

wa0x6e (approved):
tAck

ChaituVR (approved):
utAck

PR #682, opened 2023-09-25
feat(flag): add functionality for flagging verifying spaces

Todmy:
- Add a new column to the spaces table with the name `flagged` and TINYINT since it only accepts 0 or 1
- Modify logic responsive for qraphQL space query

- [x] Migrate DB
```sql
ALTER TABLE spaces 
ADD COLUMN flagged INT NOT NULL DEFAULT 0,
ADD INDEX flagged (flagged), ALGORITHM=INPLACE;
``` 

_Notes:_
the column `verified` was added earlier

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,9 +21,9 @@
   "prettier": "@snapshot-labs/prettier-config",
   "dependencies": {
     "@snapshot-labs/keycard": "^0.4.0",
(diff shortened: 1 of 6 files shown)


wa0x6e (approved):
tAck

ChaituVR (approved):
utAck

PR #687, opened 2023-10-01
fix: remove unused delegation column in proposals

wa0x6e:
Following https://github.com/snapshot-labs/snapshot-sequencer/pull/181

Remove any references to the `delegation` column if the proposals table

Also drop the column with the following query 

```sql
ALTER TABLE proposals DROP COLUMN delegation;
```


ChaituVR:
also should remove index right

```SQL
DROP INDEX delegation on proposals;

ALTER TABLE proposals
DROP COLUMN delegation;
```

ChaituVR (approved):
utAck

PR #688, opened 2023-10-01
fix: Remove delegation column from proposals table

ChaituVR:
We are using this,

Make sure to run following on prod DB and testnet DB

- [ ]  RUN

```SQL
DROP INDEX delegation on proposals;

ALTER TABLE proposals
DROP COLUMN delegation;
```


ChaituVR:
oops! duplicate of https://github.com/snapshot-labs/snapshot-hub/pull/687

PR #689, opened 2023-10-02
feat(graphql): add `address` to where query for messages

Todmy:
- change graphql schema
- it provides functionality for search address in messages


ChaituVR (approved):
utAck

PR #697, opened 2023-10-04
fix: Return error inside Promise.reject

ChaituVR:
## Why? 

- https://discord.com/channels/955773041898573854/1030772407226601522/1158324189103661096
- 
GraphQl's response without `new Error`:
```
"message": "Unexpected error value: \"address is required\"",
```

with `new Error`:
```
"message": "address is required",
```


Todmy (approved):
utACK

PR #700, opened 2023-10-07
feat: add hibernating column to Space

wa0x6e:
Add column `hibernating` to Space

```sql
ALTER TABLE spaces 
ADD COLUMN hibernating INT NOT NULL DEFAULT 0,
ADD INDEX hibernating (hibernating), ALGORITHM=INPLACE;
```

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -384,6 +384,7 @@ type Space {
   template: String
   verified: Boolean
   flagged: Boolean
+  hibernating: Boolean
   rank: Float
 }
 
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -92,7 +92,7 @@ function mapSpaces() {
 
 async function loadSpaces() {
   const query =
-    'SELECT id, settings, flagged, verified FROM spaces WHERE deleted = 0 ORDER BY id ASC';
+    'SELECT id, settings, flagged, verified, hibernating FROM spaces WHERE deleted = 0 ORDER BY id ASC';
   const s = await db.queryAsync(query);
   spaces = Object.fromEntries(
     s.map(ensSpace => [
@@ -100,7 +100,8 @@ async function loadSpaces() {
       {
         ...JSON.parse(ensSpace.settings),
         flagged: ensSpace.flagged === 1,
-        verified: ensSpace.verified === 1
+        verified: ensSpace.verified === 1,
+        hibernating: ensSpace.verified === 1
       }
     ])
   );


src/helpers/spaces.ts
```diff
@@ -92,15 +92,16 @@ function mapSpaces() {
...
   const s = await db.queryAsync(query);
   spaces = Object.fromEntries(
     s.map(ensSpace => [
       ensSpace.id,
       {
         ...JSON.parse(ensSpace.settings),
         flagged: ensSpace.flagged === 1,
-        verified: ensSpace.verified === 1
+        verified: ensSpace.verified === 1,
+        hibernating: ensSpace.verified === 1
```
bonustrack, line 104: What is this for?

ChaituVR (changes requested):
Not changing sql file? 🤔

src/graphql/schema.gql
ChaituVR, line 1: we should also add `space

src/graphql/helpers.ts
```diff
@@ -154,7 +154,7 @@ export function buildWhereQuery(fields, alias, where) {
 export async function fetchSpaces(args) {
   const { first = 20, skip = 0, where = {} } = args;
 
-  const fields = { id: 'string' };
+  const fields = { id: 'string', created_at: 'number' };
```
ChaituVR, line 157: we will pass exact timestamp to get a space?? why do we need this??

Also `created_at` is not added in `src/graphql/schema.gql` for inputs type of `Space` query, so it will throw error if we pass `created_at`

ChaituVR (changes requested):

wa0x6e:
> Not changing sql file? 🤔

fixed

src/graphql/helpers.ts
wa0x6e: column removed, will find another way from service to filter by date

src/helpers/spaces.ts
```diff
@@ -89,15 +89,16 @@ function mapSpaces() {
 
 async function loadSpaces() {
   const query =
-    'SELECT id, settings, flagged, verified FROM spaces WHERE deleted = 0 ORDER BY id ASC';
```
ChaituVR, line 92: we need to add this field to the `proposals`, `proposal`, and `votes` methods too, like here https://github.com/snapshot-labs/snapshot-hub/pull/709/

ChaituVR (changes requested):

samuveth:
When and how is this hibernation updated? @wa0x6e

samuveth:
Please link github pitches from https://github.com/snapshot-labs/pitches to make things easier (changed it for you already), we are full github mode

wa0x6e:
> When and how is this hibernation updated? @wa0x6e

From internal discussion, script will be run manually once, and filters should be put on sequencer to prevent future space from misconfiguration.

script will also be run manually as needed

wa0x6e:
> Please link github pitches from https://github.com/snapshot-labs/pitches to make things easier (changed it for you already), we are full github mode

True, pr was opened a month ago, at the time when pitches were still on notion

src/graphql/schema.gql
```diff
@@ -384,6 +384,7 @@ type Space {
   template: String
   verified: Boolean
   flagged: Boolean
+  hibernated: Boolean
```
samuveth, line 387: The issue is with this boolean approach is that we will have to duplicate the hibernation checking logic on the frontend to be able to determine the reason for hibernation and show the right information to the user.

Also what if the users just wanted to change the network from testnet to mainnet in order to de-hibernate their space? This would require manual action from us again?

@ChaituVR @wa0x6e

samuveth: To be honest I liked the https://github.com/wa0x6e/morpheus approach better

samuveth: In care we would go with the PR we could show a generic message that the space has been hibernated and that they need to open a ticket on Discord in order to get it resolved

wa0x6e: > The issue is with this boolean approach is that we will have to duplicate the hibernation checking logic on the frontend to be able to determine the reason for hibernation and show the right information to the user.
> 
> Also what if the users just wanted to change the network from testnet to mainnet in order to de-hibernate their space? This would require manual action from us again?
> 
> @ChaituVR @wa0x6e

Switching between mainnet and testnet will not be possible soon

samuveth: > Switching between mainnet and testnet will not be possible soon

Right, because sequencer will just reject. We should disable all settings on UI too then

samuveth: > We could show a link to the docs, listing reasons. And there, we can add note to go to discord

@zzuziak could you come up with a flow docs > discord > support flow for de-hibernation 

I will add a message with link to docs on UI

samuveth: @wa0x6e one more issue is though that we can not change space settings. So if a space has a testnet and they request to de-hibernate a space on discord what do we do?

wa0x6e: > @wa0x6e one more issue is though that we can not change space settings. So if a space has a testnet and they request to de-hibernate a space on discord what do we do?

They should ensure their space conform to the requirement before asking for reactivation

samuveth: As you said `Switching between mainnet and testnet will not be possible soon`

wa0x6e: We should forbid only proposal creation for hibernated spaces.

Also, people should be able to switch between mainnet and testnet only if they’re on the wrong realm

samuveth: But that's what I said, and you said it will not be possible

wa0x6e: Ok, so the new workflow from this new hibernation implementation will be:

- Spaces will be hibernated from the backend
- Hibernated space will not be able to create new proposals anymore
- There's 2 main reasons for hibernation: inactivity, and misconfiguration
- For misconfigured spaces, user will have to fix the space settings, then reapply for reactivation from discord

Misconfigutation is often wrong network, or missing validation.

For wrong network (livenet/mainnet), it's always due to users being on testnet, while being on snapshot.org, where testnet support will be dropped. 
In the space network dropdown, all testnet will be removed, beside the one the space is on (if on testnet), once he selected another option, and save, testnet will disappear from dropdown.

wa0x6e: How do we know if a space if inactive of misconfigured, to know if we should reactivate ? We can't without a service that tells you that. DB column is a simple boolean

samuveth: Agree to all, except that the dropdown input will be empty if the network is from the wrong env

> In the space network dropdown, all testnet will be removed, beside the one the space is on (if on testnet), once he selected another option, and save, testnet will disappear from dropdown.

samuveth: image

wa0x6e: That also works

ChaituVR (dismissed):
utAck

ChaituVR (approved):
tAck

PR #702, opened 2023-10-09
fix: upgrade snapshot.js to 0.8.0 beta.0

### LINKED ISSUE #413: Don't return error object in response

wa0x6e:
Upgrade snapshot.js to 0.8.0-beta.0, with `ofetch` migration.

- Add try/catch around snapshot.utils function, and return a more meaningful error
- Return error when space or proposal are not found
- Return error when voter is blank (ideally, we should check that it's a valid address here)

There's only one file affected: operation/vp.ts

Should also fix #413 

Before upgrade, error response

```json
{
  "errors": [
    {
      "message": "Unexpected error value: { code: 500, message: \"unauthorized\", data: {} }",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "vp"
      ]
    }
  ],
  "data": {
    "vp": null
  }
}
```

After upgrade, error response:

```json
{
  "errors": [
    {
      "message": "request failed",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "vp"
      ]
    }
  ],
  "data": {
    "vp": null
  }
}
```

## Test

Working query 

```
query Vp{
  vp(space: "stgdao.eth", voter:"0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3") {
    vp_state
    vp_by_strategy
  }
}
```

Failing query

```
query Vp{
  vp(space: "magicappstore.eth", voter:"0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3") {
    vp_state
    vp_by_strategy
  }
}
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
     "@snapshot-labs/keycard": "^0.4.0",
     "@snapshot-labs/snapshot-metrics": "^1.3.1",
     "@snapshot-labs/snapshot-sentry": "^1.5.1",
-    "@snapshot-labs/snapshot.js": "^0.7.2",
+    "@snapshot-labs/snapshot.js": "^0.8.0-beta.0",
     "bluebird": "^3.7.2",
     "connection-string": "^1.0.1",
     "cors": "^2.8.5",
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
tAck

PR #703, opened 2023-10-09
fix: return error when space does not exist

wa0x6e:
Fixes #478 and #413 

Return a 404 error when the space is not found


ChaituVR:
It won't fix https://github.com/snapshot-labs/snapshot-hub/issues/413 - it needs more refactoring for example we still return error object in some cases which contain SQL information, example https://discord.com/channels/707079246388133940/1160775468845387776/1160775771216953415

ChaituVR (approved):
utAck

PR #704, opened 2023-10-10
feat: filter spaces by `strategies`, `plugins` and `network`

### LINKED ISSUE #535: Query to get spaces using a plugin
https://discord.com/channels/707079246388133940/747049240022482974/1057595431514816572

### LINKED ISSUE #548: Add strategies_contain and plugins_contain to space related queries
Can we also add that query to `Spaces` or is it only possible at a proposal level?

Ideally, it would be great to retrieve a list of spaces using SafeSnap, POAP, etc.

_Originally posted by @zzuziak in https://github.com/snapshot-labs/snapshot-hub/issues/546#issuecomment-1425988607_

wa0x6e:
Fix #548 
Fix #535 

Add `strategies_in` and `plugins_in` filter to Space.

This allow query such as:

```graphql
query Spaces {
  spaces(
    first: 10,
    where: { plugins_in: ["poap", "progress"], strategies_in: ["whitelist"], network: "1" }
  ) {
    strategies {
      name
    }
    plugins
    network
  }
}
```

Those filters behave differently than the proposal's `strategies_contains`. 

`strategies_contains` is using string search, and can match anything, such as `{`.
`strategies/plugins_in` filter will exact match the strategy/plugin name.

Feature has been applied to Proposal via https://github.com/snapshot-labs/snapshot-hub/pull/705

The strategies search only apply to level 2 strategies, and will not match nested strategies

Creating the following indices will improve search performance

```sql

ALTER TABLE spaces
ADD INDEX strategies_idx ((CAST(JSON_EXTRACT(settings, "$.strategies[*].name") AS CHAR(255) ARRAY) )) USING BTREE;

ALTER TABLE spaces
ADD INDEX plugins_idx ((CAST(JSON_KEYS(settings, "$.plugins") AS CHAR(255) ARRAY) )) USING BTREE;

ALTER TABLE spaces
ADD INDEX network_idx ((CAST(JSON_UNQUOTE(JSON_EXTRACT(settings, "$.network")) AS CHAR(255)) COLLATE utf8mb4_bin )) USING BTREE;
```

The goal of those new filters will be to enable the snapshot UI to show the list of spaces using specific network/strategy/plugin

## Note 

`JSON_OVERLAPS` require Mysql version >= 8.0.17

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -156,7 +156,7 @@ export async function fetchSpaces(args) {
 
   const fields = { id: 'string' };
   const whereQuery = buildWhereQuery(fields, 's', where);
-  const queryStr = whereQuery.query;
+  let queryStr = whereQuery.query;
   const params: any[] = whereQuery.params;
 
   let orderBy = args.orderBy || 'created_at';
@@ -164,6 +164,16 @@ export async function fetchSpaces(args) {
   if (!['created_at', 'updated_at', 'id'].includes(orderBy)) orderBy = 'created_at';
   orderDirection = orderDirection.toUpperCase();
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+
+  if (where.strategies_in) {
+    queryStr += " AND JSON_OVERLAPS(JSON_EXTRACT(settings, '$.strategies[*].name'), JSON_ARRAY(?))";
+    params.push(where.strategies_in);
+  }
+
+  if (where.plugins_in) {
+    queryStr += " AND JSON_OVERLAPS(JSON_KEYS(settings, '$.plugins') , JSON_ARRAY(?))";
+    params.push(where.plugins_in);
+  }
 
   const query = `
     SELECT s.* FROM spaces s
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -175,6 +175,11 @@ export async function fetchSpaces(args) {
     params.push(where.plugins_in);
   }
 
+  if (where.network) {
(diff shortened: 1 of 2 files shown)


src/graphql/helpers.ts
```diff
@@ -165,6 +165,21 @@ export async function fetchSpaces(args) {
   orderDirection = orderDirection.toUpperCase();
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
 
+  if (where.strategies_in) {
+    queryStr += " AND JSON_OVERLAPS(JSON_EXTRACT(settings, '$.strategies[*].name'), JSON_ARRAY(?))";
```
ChaituVR, line 169: It only returns spaces that return space that uses only one strategy, it should return any space that use the strategy being passed

For example if space A is using `ticket` and `erc721` and space B is using just `ticket` strategy,
now if we request with `strategies_in:["ticket"]` it return only space B, it should return space A too

ChaituVR (changes requested):

src/graphql/helpers.ts
wa0x6e: It should be the case already, that's the job of `JSON_OVERLAPS`. Do you have a specific example I can test ? 

Running the query in the description is giving me spaces using multiple strategies

```graphql
"strategies": [
          {
            "name": "whitelist"
          },
          {
            "name": "delegation"
          },
          {
            "name": "moloch-all"
          },
          {
            "name": "loot-character-guilds"
          },
          {
            "name": "rocketpool-node-operator"
          },
          {
            "name": "honeyswap"
          },
          {
            "name": "xdai-stake-holders"
          },
          {
            "name": "yearn-vault"
          }
        ],
```

PR #707, opened 2023-10-11
fix: instrument graphql requests

wa0x6e:

Current graphql requests are not instrumented

Will answer this kind of question: https://discord.com/channels/955773041898573854/1103266859870081114/1161683893968183367

## 💊 Fixes / Solution

Instrument the graphql requests, to extract the number of requests and response time per type (Space, Proposals, Votes, etc...)

## 🚧 Changes

- Add middleware to instrument the graphql requests

## 🛠️ Tests

- Start the service
- Send some requests to http://localhost:3000/graphql
- You should see metrics with the key `graphql_requests_duration_seconds_bucket `

### DIFF
--- a/src/helpers/metrics.ts
+++ b/src/helpers/metrics.ts
@@ -1,9 +1,11 @@
 import init, { client } from '@snapshot-labs/snapshot-metrics';
 import { capture } from '@snapshot-labs/snapshot-sentry';
-import { Express } from 'express';
+import { Express, type Request, type Response } from 'express';
+import { parse } from 'graphql';
 import { spacesMetadata } from './spaces';
 import { strategies } from './strategies';
 import db from './mysql';
+import operations from '../graphql/operations/';
 
 const whitelistedPath = [
   /^\/$/,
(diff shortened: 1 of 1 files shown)


src/helpers/metrics.ts
```diff
@@ -41,6 +45,39 @@ export default function initMetrics(app: Express) {
   });
 
   app.use(instrumentRateLimitedRequests);
+
+  app.use((req: Request, res: Response, next) => {
+    if (req.originalUrl.startsWith('/graphql')) {
+      const end = graphqlRequest.startTimer();
```
ChaituVR, line 51: We are not ending this if there is no query or operationName 
```javascript
if (query && operationName) {
```

ChaituVR (changes requested):

src/helpers/metrics.ts
wa0x6e: We are tracking only valid queries

ChaituVR: oh i thought it would stay in memory for ever 🤔

ChaituVR (approved):
utAck

PR #710, opened 2023-10-12
fix: Move from graphql-tools to @graphql-tools/schema

ChaituVR:
We get the following notice
```bash
This package has been deprecated and now it only exports makeExecutableSchema.
And it will no longer receive updates.
We recommend you to migrate to scoped packages such as @graphql-tools/schema, @graphql-tools/utils and etc.
Check out https://www.graphql-tools.com to learn what package you should use instead!
```
so moved to this new package, and everything looks to be working fine

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
   },
   "prettier": "@snapshot-labs/prettier-config",
   "dependencies": {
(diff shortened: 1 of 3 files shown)


wa0x6e (approved):
tAck

PR #712, opened 2023-10-14
feat: filter votes by choice

wa0x6e:
Another implementation of #571 

Add `choice_in` filter to Vote, allowing query such as 

```
query Vote {
  votes(where: {choice_in: [15]}) {
    id
    choice
  }
}
```

## Other changes

- Refactor moderation and spaces cache to run only when called, for easier tests

## Notes

It is impossible to add an INDEX on the choice JSON column, as the stored data format is not consistent, and can be either an INT, an array of INT, or an Object.

The filter should ideally be used only along some other filter, such as proposal id, etc ... to filter down the results first

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -9,6 +9,7 @@ import log from './helpers/log';
 import initMetrics from './helpers/metrics';
 import { checkKeycard } from './helpers/keycard';
 import refreshModerationData from './helpers/moderation';
+import refreshSpacesCache from './helpers/spaces';
 import './helpers/strategies';
 
 const app = express();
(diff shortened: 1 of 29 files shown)


src/index.ts
```diff
@@ -8,14 +8,15 @@ import rateLimit from './helpers/rateLimit';
...
-import './helpers/moderation';
+import refreshModerationData from './helpers/moderation';
 import './helpers/strategies';
 
 const app = express();
 const PORT = process.env.PORT || 3000;
 
 initLogger(app);
 initMetrics(app);
+refreshModerationData();
```
ChaituVR, line 19: Ideally we should avoid other changes on PRs with different context, 😄 when i look at PR it contains 
image

It scares me, so i add it to my todo and get delayed 🙈

wa0x6e: PR is less scary now, majority of changes are tests and config files

PR #713, opened 2023-10-15
chore: ensure test suite are only run on test database

wa0x6e:
This PR add additional securities to ensure test suite are only run on test database.

The convention will be to always prefix test database with `_test`. `yarn test:setup` is provided to automatically setup the test database


ChaituVR (approved):
utAck

PR #716, opened 2023-10-18
fix: get `skins` data from DB directly

wa0x6e:

The `skin` query on the hub is using data from space's cache.

## 💊 Fixes / Solution

Use a SQL query to get skins data from the database

## 🚧 Changes

- Get skins data from the database

## 🛠️ Tests

- Query some skin data from graphql => it should return same results as before


wa0x6e:
Query is taking between 300 and 350ms to execute on main net database.

This is reasonable, considering a simple `SELECT id FROM spaces WHERE id = ?` is taking between 250 and 300ms.

wa0x6e:
@ChaituVR reopening this one, as it's now relevant again in our quest of removing the big global `spaces` object

wa0x6e:
Furthermore, this query is now rarely used, since only used on v1, v2 has a different skin system

wa0x6e:
`skins` is now dead, as it's only used on v1, which does not support whitelabel anymore.

PR #717, opened 2023-10-18
fix: get `validation` data from DB directly

wa0x6e:

The `validation` query on the hub is using data from space's cache.

## 💊 Fixes / Solution

Use a SQL query to get validations data from the database

## 🚧 Changes

- Get validations data from the database

## 🛠️ Tests

- Query some validation data from graphql => it should return same results as before


wa0x6e:
Pretty fast query, taking around 300ms

PR #718, opened 2023-10-18
fix: get `strategies` data from DB directly

wa0x6e:

The `strategies` query on the hub is using data from space's cache.

## 💊 Fixes / Solution

Use a SQL query to get strategies data from the database

## 🚧 Changes

- Get strategies data from the database
- Remove the `strategiesObj` global object. The data can already be queried using a simple `find` on the `strategies`  global variable.

## 🛠️ Tests

- Query some strategies and strategy data from graphql => it should return same results as before

### DIFF
--- a/src/helpers/strategies.ts
+++ b/src/helpers/strategies.ts
@@ -1,11 +1,10 @@
 import snapshot from '@snapshot-labs/snapshot.js';
-import { spaces } from './spaces';
-import log from './log';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 import { URL } from 'url';
+import db from '../helpers/mysql';
+import log from './log';
 
 export let strategies: any[] = [];
-export let strategiesObj: any = {};
 
 const scoreApiURL: URL = new URL(process.env.SCORE_API_URL ?? 'https://score.snapshot.org');
 scoreApiURL.pathname = '/api/strategies';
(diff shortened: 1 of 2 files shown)


src/helpers/strategies.ts
```diff
@@ -21,24 +20,31 @@ async function loadStrategies() {
...
-    ids.forEach(id => {
-      if (res[id]) {
-        res[id].spacesCount = (res[id].spacesCount || 0) + 1;
-      }
-    });
-  });
+  const results = new Map(
+    (
+      await db.queryAsync(`
+        SELECT
```
bonustrack, line 26: This can be an heavy request isn't? How long does it take to resolve? How often is it called?

wa0x6e: ~40ms to resolve, on my dataset of 13k spaces.

It's still not 100% live query, as it's only run along the fetch strategies query from score url, so once every 1min

bonustrack: Ok

wa0x6e: Do you want more refactoring, and use 100% live counters (and not cached for 1min) ? 

This will bring more consistency, as no other counters are cached and lagging behind.

By enforcing only known strategies when submitting via sequencer, we could also get rid of the condition, checking if the strategy is a valid one from score-api, and have true live data from the strategies operation. (can also get rid of the /helpers/startegies.ts file)

bonustrack: I prefer not using such query as live query, also not sure how you wouod get ride of strategies helpers, this is loading strategies metadata that are not available from the hub

wa0x6e: true, forgot about the other strategies data. We get less than 1 request/min on the hub for the strategies object, should have very minimal impact on performance

ChaituVR: Takes around 2.8s on prod, will increase when we get more space 🤔

wa0x6e:
Return to WIP, will add some cache and update strategy

PR #721, opened 2023-10-18
fix: get `plugins` data from DB directly

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/544

The `plugins` query on the hub is using data from space's cache.

## 💊 Fixes / Solution

Use a SQL query to get plugins data from the database

## 🚧 Changes

- Get plugins data from the database

## 🛠️ Tests

- Query some plugin data from graphql => it should return same results as before


wa0x6e:
Pretty fast query, taking around 350ms

PR #722, opened 2023-10-18
fix: get `network` data from DB directly

wa0x6e:

The `network` query on the hub is using data from space's cache.

## 💊 Fixes / Solution

Use a SQL query to get networks count data from the database, and return more network properties, allowing queries like:

```graphql
query Networks {
  networks {
    key
    id
    spacesCount
    name
    chainId
    network
    explorer {
      url
      apiUrl
    }
    logo
    testnet
  }
}
```

## 🚧 Changes

- Get networks count data from the database instead of space's cache
- Return more fields from networks.json. Not all fields from networks.json are returned, only the one used by the snapshot UI.

## 🛠️ Tests

- Query some network data from graphql => it should return same results as before, in addition to the new fields

### DIFF
--- a/src/graphql/operations/networks.ts
+++ b/src/graphql/operations/networks.ts
@@ -1,12 +1,20 @@
-import { spaces } from '../../helpers/spaces';
+import { capture } from '@snapshot-labs/snapshot-sentry';
+import db from '../../helpers/mysql';
 
-export default function () {
-  const networks = {};
-  Object.values(spaces).forEach((space: any) => {
-    networks[space.network] = networks[space.network] ? networks[space.network] + 1 : 1;
-  });
-  return Object.entries(networks).map(network => ({
-    id: network[0],
-    spacesCount: network[1]
-  }));
+export default async function () {
+  const query = `
+    SELECT
+      JSON_UNQUOTE(JSON_EXTRACT(settings, '$.network')) as network,
+      COUNT(name) AS spacesCount
+    FROM spaces
+    GROUP BY network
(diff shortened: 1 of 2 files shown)


wa0x6e:
> But then we will miss new networks , where no one is using them yet

Aren’t they already missing from current implementation ?

ChaituVR:
> Aren’t they already missing from current implementation ?

True lol. i thought we were using `networks.json`,

src/graphql/operations/networks.ts
```diff
@@ -1,12 +1,20 @@
...
+export default async function () {
+  const query = `
+    SELECT
+      JSON_UNQUOTE(JSON_EXTRACT(settings, '$.network')) as network,
+      COUNT(name) AS spacesCount
+    FROM spaces
+    GROUP BY network
+    ORDER BY spacesCount DESC;
+  `;
+
```
ChaituVR, line 13: Takes around 0.5 sec on prod, should be fine?

wa0x6e:
> Would be nice if we could add it now 😄 so then maybe we can remove networks.json dependency on snapshot UI

Added

wa0x6e:
Not all fields from networks.json are returned, tell me if we need to add more.

wa0x6e:
Pretty fast sql query, averaging 350ms execution time.

We could save the computation by caching the whole thing, and using `SELECT MAX(updated) FROM spaces;` as cache key

wa0x6e:
Need fix to also include networks from inside strategies

PR #725, opened 2023-10-19
fix: `/api/spaces/ID` should pull data from DB instead of cache

wa0x6e:

The `/api/spaces/:SPACE_ID` endpoint is pulling data from the space's cache.

It's not using any additional data set by the cache (such as metrics), and data can be pulled directly from the database using a simple `SELECT` query, bypassing the cache and it's latency, and reducing side effect by using the cache.

Using the cache is making the code impossible to tests

## 💊 Fixes / Solution

- Use space data from the database directly

## 🚧 Changes

- Use data from database directly
- Add some tests

## 🛠️ Tests

- get `/api/spaces/[ID]` => it should return same data as before

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,10 +55,10 @@
     "eslint": "^8.28.0",
     "jest": "^29.6.1",
     "jest-environment-node-single-context": "^29.1.0",
+    "node-fetch": "^2.7",
     "nodemon": "^2.0.19",
     "prettier": "^3.0.3",
     "start-server-and-test": "^2.0.0",
-    "supertest": "^6.3.3",
     "ts-jest": "^29.1.1"
   },
   "engines": {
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -55,10 +55,10 @@
     "eslint": "^8.28.0",
     "jest": "^29.6.1",
     "jest-environment-node-single-context": "^29.1.0",
+    "node-fetch": "^2.7",
```
ChaituVR, line 58: why this change? 🤔

ChaituVR, line 58: Maybe we should  use `ofetch` package?

src/helpers/spaces.ts
```diff
@@ -158,6 +158,24 @@ async function loadSpacesMetrics() {
   mapSpaces();
 }
 
+export async function getSpace(id: string) {
+  const query = `
+    SELECT id, settings, flagged, verified
```
ChaituVR, line 163:
```suggestion
    SELECT settings, flagged, verified
```

ChaituVR (approved):
tAck

PR #726, opened 2023-10-19
fix: fix space query unable to get past 30k spaces

### LINKED ISSUE #714: Can't get spaces after 30k
### Expected behavior
The Graphql query should return further spaces 

### Actual behavior
But we can't skip parameter > 30k
image

### Steps to reproduce the behavior
```
query space_data {spaces(first:10,skip:31000){id}}
```

wa0x6e:
Fix #714 

We're limiting the space to 30k, which is a hard limit, since the Space query does not provided any filters to paginate.

This PR adds a `created` filter in where, to enable paginating the Space object, like the Vote object.

This enable query like 

```graphql
query Spaces {
  spaces(first: 10, skip: 5, where: {created_gt: 1697661410}) {
    name
    created
  }
}
```

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -117,12 +117,12 @@ type Query {
 input SpaceWhere {
   id: String
   id_in: [String]
-  created_at: Int
-  created_at_in: [Int]
-  created_at_gt: Int
-  created_at_gte: Int
-  created_at_lt: Int
-  created_at_lte: Int
+  created: Int
+  created_in: [Int]
+  created_gt: Int
+  created_gte: Int
+  created_lt: Int
+  created_lte: Int
 }
 
 input RankingWhere {
@@ -391,7 +391,7 @@ type Space {
   verified: Boolean
   flagged: Boolean
   rank: Float
-  created_at: Int!
+  created: Int!
 }
 
 type RankingObject {
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -117,6 +117,12 @@ type Query {
 input SpaceWhere {
   id: String
   id_in: [String]
+  created_at: Int
+  created_at_in: [Int]
+  created_at_gt: Int
+  created_at_gte: Int
+  created_at_lt: Int
+  created_at_lte: Int
 }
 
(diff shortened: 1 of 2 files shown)


src/graphql/schema.gql
```diff
@@ -385,6 +391,7 @@ type Space {
   verified: Boolean
   flagged: Boolean
   rank: Float
+  created_at: Int!
```
ChaituVR, line 394: filters like created_at_gte or created_at_lte are more useful to filter, created_at is just a single timestamp.  so it will be hard to get more spaces than 30k limit

wa0x6e: Filters are in `SpaceWhere` (https://github.com/snapshot-labs/snapshot-hub/pull/726/files#diff-f5d6a46a522635fe64ba4ec435a92609da73330c512801e8c0efd1cd7807a8b0R120), this is just the returned value

ChaituVR: This confused me 😆 
Untitled 2

ChaituVR (approved):
utAck

src/graphql/schema.gql
ChaituVR, line 394: Maybe we have rename this field to `created` and `updated_at` like all other tables first???

wa0x6e: This change needs some more planification, as it may breaks the UI and other services using that field, unless we can alias both field in a transition period

ChaituVR: > as it may breaks the UI and other services using that field

How other services or UI using this field when this PR is the first time we adding this field 😅

wa0x6e: Valid point

wa0x6e: In that case, we should also rename the `updated_at` field in Space, for consistency.

In another note, the `_at` suffix for timestamp property make more sense, as it differentiate if from boolean field (`verified`), and it's also the convention used by some CMS, like Ruby on Rails and Symfony.

wa0x6e: done

ChaituVR (approved):
utAck

PR #730, opened 2023-10-25
fix: check for value before manipulation

### LINKED ISSUE #729: TypeError: Cannot read properties of undefined (reading 'selectionSet')
Sentry Issue: [SNAPSHOT-HUB-1B](https://snapshot-labs.sentry.io/issues/4548600448/?referrer=github_integration)

```
TypeError: Cannot read properties of undefined (reading 'selectionSet')
  File "/home/node/app/dist/src/helpers/metrics.js", line 81, in ServerResponse.<anonymous>
    const types = definition.selectionSet.selections.map(sel => sel.name.value);
...
(3 additional frame(s) were not displayed)
```

wa0x6e:
Fix #729
Fix SNAPSHOT-HUB-1D
Fix SNAPSHOT-HUB-1B

Check that value is defined before usage

### DIFF
--- a/src/helpers/metrics.ts
+++ b/src/helpers/metrics.ts
@@ -58,8 +58,12 @@ export default function initMetrics(app: Express) {
           if (query && operationName) {
             const definition = parse(query).definitions.find(
               // @ts-ignore
-              def => def.name.value === operationName
+              def => def.name?.value === operationName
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

PR #731, opened 2023-10-27
refactor: trigger data refresher manually

wa0x6e:
Extracted from #712 

Trigger the space and moderation data refresher manually, instead of automatically on import.

This allow skipping the refresh, and is particularly useful on test suite

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,14 +1,12 @@
 import express from 'express';
 import { capture } from '@snapshot-labs/snapshot-sentry';
-import refreshSpacesCache, { getSpace } from './helpers/spaces';
+import { getSpace } from './helpers/spaces';
 import { name, version } from '../package.json';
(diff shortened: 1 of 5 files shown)


src/api.ts
```diff
@@ -1,13 +1,15 @@
...
-import { getSpace } from './helpers/spaces';
+import refreshSpacesCache, { getSpace } from './helpers/spaces';
 import { name, version } from '../package.json';
 import { sendError } from './helpers/utils';
 
 const router = express.Router();
 const network = process.env.NETWORK || 'testnet';
 const SEQUENCER_URL = process.env.SEQUENCER_URL || '';
 
+refreshSpacesCache();
```
ChaituVR, line 11: Why this one here and `refreshModerationData` in index.ts 😅 better to keep all in one place

ChaituVR (approved):
utAck

PR #739, opened 2023-11-04
chore: instrument flagged spaces count

wa0x6e:
- Add metric to track the number of flagged spaces
- Remove the unverified spaces metrics (already can be computed from total_spaces - verified_spaces)


ChaituVR (approved):
utAck

Todmy (approved):
utACK

PR #740, opened 2023-11-04
fix: rename `_at` suffix from Space timestamp properties

wa0x6e:
Ref to #726 

Rename the `created_at` and `updated_at` properties in Space to just `created` and `updated`, to be consistent with the all the other objects.

Run the following query 

```sql
ALTER TABLE spaces RENAME COLUMN created_at TO created;
ALTER TABLE spaces RENAME INDEX created_at TO created;

ALTER TABLE spaces RENAME COLUMN updated_at TO updated;
ALTER TABLE spaces RENAME INDEX updated_at TO updated;
```

### DIFF
--- a/src/helpers/schema.sql
+++ b/src/helpers/schema.sql
@@ -6,14 +6,14 @@ CREATE TABLE spaces (
   deleted INT NOT NULL DEFAULT '0',
   flagged INT NOT NULL DEFAULT '0',
   created BIGINT NOT NULL,
-  updated_at BIGINT NOT NULL,
+  updated BIGINT NOT NULL,
   PRIMARY KEY (id),
   INDEX name (name),
   INDEX verified (verified),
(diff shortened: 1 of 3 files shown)


src/helpers/schema.sql
```diff
@@ -5,15 +5,15 @@ CREATE TABLE spaces (
   verified INT NOT NULL DEFAULT '0',
   deleted INT NOT NULL DEFAULT '0',
   flagged INT NOT NULL DEFAULT '0',
-  created_at BIGINT NOT NULL,
-  updated_at BIGINT NOT NULL,
+  created BIGINT NOT NULL,
```
ChaituVR, line 8: Should we also default these to `DEFAULT (UNIX_TIMESTAMP())`

ChaituVR (approved):
utAck

PR #742, opened 2023-11-06
fix: Remove confusing documentation

ChaituVR:
the text is confusing, someone using hub doesn't have to provide URL to us, it should run individually


wa0x6e (approved):
utAck

PR #750, opened 2023-11-13
Fix: Exclude flagged proposals from active proposals count

ChaituVR:
Fixes https://github.com/snapshot-labs/pitches/issues/47

How to test:
- Create a test proposal on your space, `activeProposals` on `Space` query should return `1`
- Mark proposal as flagged
- Count should be `0`


Todmy (approved):
tACK

PR #763, opened 2023-11-28
fix: empty array in graphql filters should trigger an error

### LINKED ISSUE #757: Params with `_in` like `proposal_in` should not be empty arrays
Right now we return a response even if the user passes empty arrays, I think we should return an error in such cases  

Example query 
```Graphql
query Votes($voter: String!, $proposals: [String]!) {
    votes(first: 1000, where: { voter: $voter, proposal_in: $proposals }) {
      proposal {
        id
        space {
          id
        }
      }
    }
  }
 
  ```
variables
```JSON
{
  "voter": "0x24F15402C6Bb870554489b2fd2049A85d75B982f",
  "proposals": []
}
```

wa0x6e:
Fixes #757 

GraphQL filters with empty array now return an error, like on https://api.lens.dev/graphql

Test with 

```graphql
query Spaces {
  spaces(first: 5, skip: 0, where: {id_in: []}) {
    name
    created
  }
}

```

Should return 

```json
{
  "errors": [
    {
      "message": "id_in must have at least one item",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "spaces"
      ]
    }
  ],
  "data": {
    "spaces": null
  }
}
```

Additionally, move all `throw new Error` to `throw new PulbicError` for consistency

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -102,6 +102,7 @@ export function formatSpace({ id, settings, verified, flagged, hibernated }) {
 export function buildWhereQuery(fields, alias, where) {
   let query: any = '';
   const params: any[] = [];
+
   Object.entries(fields).forEach(([field, type]) => {
     if (where[field] !== undefined) {
       query += `AND ${alias}.${field} = ? `;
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
tAck

PR #764, opened 2023-11-29
Revert "fix: empty array in graphql filters should trigger an error"

ChaituVR:
Reverts snapshot-labs/snapshot-hub#763

Looks like UI still send empty array sometimes, we need to fix those first 

For example in proposals page
https://discord.com/channels/707079246388133940/1090304570015629365/1179358781788913674


wa0x6e (approved):
UtAck

PR #765, opened 2023-11-29
fix: return empty results on empty array filters

### LINKED ISSUE #757: Params with `_in` like `proposal_in` should not be empty arrays

wa0x6e:
Fixes #757 

Return an empty response when array filter is empty
Retry of https://github.com/snapshot-labs/snapshot-hub/pull/763 which was reverted

Test with 

```graphql
query Spaces {
  spaces(first: 5, skip: 0, where: {id_in: []}) {
    name
    created
  }
}

```

Should return 

```json
{
  "errors": [
    {
      "message": "id_in must have at least one item",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "spaces"
      ]
    }
  ],
  "data": {
    "spaces": null
  }
}
```

Additionally, move all `throw new Error` to `throw new PulbicError` for consistency

# TODO

- [ ] Merge https://github.com/snapshot-labs/snapshot/pull/4418 first

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -102,6 +102,7 @@ export function formatSpace({ id, settings, verified, flagged, hibernated }) {
 export function buildWhereQuery(fields, alias, where) {
   let query: any = '';
   const params: any[] = [];
+
   Object.entries(fields).forEach(([field, type]) => {
     if (where[field] !== undefined) {
       query += `AND ${alias}.${field} = ? `;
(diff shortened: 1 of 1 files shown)


src/graphql/helpers.ts
```diff
@@ -114,16 +115,24 @@ export function buildWhereQuery(fields, alias, where) {
       params.push(fieldNot);
     }
 
-    const fieldIn = where[`${field}_in`] || [];
-    if (fieldIn.length > 0) {
-      query += `AND ${alias}.${field} IN (?) `;
-      params.push(fieldIn);
+    const fieldIn = where[`${field}_in`];
+    if (Array.isArray(fieldIn)) {
```
ChaituVR, line 119: Looks like `_in`s are always array, we don't need to check

even if we query
```
query Votes {
    spaces(first: 1000, where: { id_in: "fabien.eth" }) {
      id
    }
  }
  ```
  
  `id_in` will be an array

Comment (ChaituVR) [reply src/graphql/helpers.ts:119]: I mean can be like before: 
```
    const fieldIn = where[`${field}_in`] || [];
```

Comment (wa0x6e) [reply src/graphql/helpers.ts:119]: > I mean can be like before:
> 
> ```
>     const fieldIn = where[`${field}_in`] || [];
> ```

`where[`${field}_in`]` can be either 
- something valid (continue), 
- an empty array (stop), 
- or undefined (ignore)

If we cast with `|| []`, example on the space query with `id` and `created` filter:

```
spaces(first: 1000, where: { id_in: "fabien.eth" })
```

will be considered as 

```
spaces(first: 1000, where: { id_in: "fabien.eth", created_in: [] })
``` 

and will always return nothing

ChaituVR: Missed this point, thank you!

PR #769, opened 2023-12-12
chore: add turbo param for spaces

bonustrack:
The field is already added in the testnet and livenet db. This will be used to grant additional permission or advantages for spaces.

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -391,6 +391,7 @@ type Space {
   verified: Boolean
   flagged: Boolean
   hibernated: Boolean
+  turbo: Boolean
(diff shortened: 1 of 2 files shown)


src/graphql/schema.gql
```diff
@@ -391,6 +391,7 @@ type Space {
   verified: Boolean
   flagged: Boolean
   hibernated: Boolean
+  turbo: Boolean
```
ChaituVR, line 394: I feel like this should be a number, we may add different levels of turbo in the future, like for example higher pricing plan or just a spam-fighting one

ChaituVR (approved):
utAck

PR #774, opened 2023-12-21
feat: turbo field is missing in all queries like votes, proposals, etc.

ChaituVR:
`turbo` is missing on all queries that return `space` and it is confusing so made it similar to other fields 

### How to test:

Try following query before and after fix, `turbo` will be `null` before the fix
```
query proposals {
    proposals {
       
        id
        ipfs
        space {
          turbo
        }
      
    }
  }

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ yarn start
 
 ### Load a space setting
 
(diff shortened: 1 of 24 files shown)


wa0x6e (approved):
utAck

PR #776, opened 2023-12-22
fix: SQL query to get space is missing turbo field

ChaituVR:
`loadSpaces` is missing turbo field so it is unable to calculate ranking of turbo spaces


bonustrack:
Can you fix the test? It failing on GitHub now

ChaituVR:
Yes looking into it, noticed it after creating PR

wa0x6e (approved):
utAck

PR #785, opened 2024-01-11
fix: fix error when passing `null` to `skip` and `first` arg

### LINKED ISSUE #783: Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your M...
Sentry Issue: [SNAPSHOT-HUB-1V](https://snapshot-labs.sentry.io/issues/4833242289/?referrer=github_integration)

```
Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULL, 10' at line 10
  File "/home/node/app/dist/src/graphql/operations/proposals.js", line 92, in Object.default_1 [as proposals]
    const proposals = await mysql_1.default.queryAsync(query, params);
...
(24 additional frame(s) were not displayed)

Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULL, 10' at line 10
  File "/home/node/app/dist/src/graphql/operations/proposals.js", line 92, in Object.default_1 [as proposals]
    const proposals = await mysql_1.default.queryAsync(query, params);
...
(24 additional frame(s) were not displayed)
```

wa0x6e:

The graphql schema is allowing query like this

```
query Proposals {
  proposals(
    first: 10,
    skip: null
  ) {
    id
  }
}
```

Where `first` and `skip` can both be null, due to usage of `Int`

This query trigger an SQL error, since it's interpreted as `ORDER BY xx ASC NULL, NULL`
 
## 💊 Fixes / Solution

Disallow usage of `null` on `skip` and `first` argument, and only accept number

Fix #783 
Fix [SNAPSHOT-HUB-1V](https://snapshot-labs.sentry.io/issues/4833242289/?referrer=github_integration)

## 🚧 Changes

- Enforce default value for `skip` and `first` on the schema level
- Remove default value on resolvers function

## 🛠️ Tests

Passing following query should not fail anymore, and should trigger a validation error instead

```
query Proposals {
  proposals(
    first: 10,
    skip: null
  ) {
    id
  }
}
```

Passing a query without `skip` for `first` should use the default value

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -174,7 +174,7 @@ export function buildWhereQuery(fields, alias, where) {
 }
 
 export async function fetchSpaces(args) {
-  const { first = 20, skip = 0, where = {} } = args;
+  const { first, skip, where = {} } = args;
 
   const fields = { id: 'string', created: 'number' };
   const whereQuery = buildWhereQuery(fields, 's', where);
--- a/src/graphql/operations/aliases.ts
+++ b/src/graphql/operations/aliases.ts
@@ -4,7 +4,7 @@ import log from '../../helpers/log';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 
 export default async function (parent, args) {
-  const { first = 20, skip = 0, where = {} } = args;
+  const { first, skip, where = {} } = args;
 
   checkLimits(args, 'aliases');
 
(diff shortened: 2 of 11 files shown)


ChaituVR (approved):
tAck

PR #786, opened 2024-01-11
fix: fix missing default when args is not coming from gql schema

wa0x6e:

Default value for `args` is missing for `fetchSpaces`, when called internally, and not via graphql.

## 💊 Fixes / Solution

Add back default args values for `first` and `skip` for `fetchSpaces`

## 🛠️ Tests

When applied with #785, the following query should now works:

```
query Space {
  space(id: "aave.eth") {
    id 
  }
}
```


ChaituVR (approved):
utAck

PR #788, opened 2024-01-11
fix: fix escape issue when address contains special characters

### LINKED ISSUE #787: Error: ER_INVALID_JSON_TEXT_IN_PARAM: Invalid JSON text in argument 2 to function json_contains: "The do...
Sentry Issue: [SNAPSHOT-HUB-1Q](https://snapshot-labs.sentry.io/issues/4804737954/?referrer=github_integration)

```
Error: ER_INVALID_JSON_TEXT_IN_PARAM: Invalid JSON text in argument 2 to function json_contains: "The document root must not be followed by other values." at position 19.
  File "/home/node/app/dist/src/graphql/operations/roles.js", line 22, in Object.default_1 [as roles]
    const data = await mysql_1.default.queryAsync(query, params);
...
(24 additional frame(s) were not displayed)

Error: ER_INVALID_JSON_TEXT_IN_PARAM: Invalid JSON text in argument 2 to function json_contains: "The document root must not be followed by other values." at position 19.
  File "/home/node/app/dist/src/graphql/operations/roles.js", line 22, in Object.default_1 [as roles]
    const data = await mysql_1.default.queryAsync(query, params);
...
(24 additional frame(s) were not displayed)
```

wa0x6e:

The role graphql query is not escaping special characters properly, making the SQL query fail in some cases.

See invalid input in [SNAPSHOT-HUB-1Q](https://snapshot-labs.sentry.io/issues/4804737954/?referrer=github_integration)

## 💊 Fixes / Solution

Close #787 
Close [SNAPSHOT-HUB-1Q](https://snapshot-labs.sentry.io/issues/4804737954/?referrer=github_integration)

Escape the `address` input properly by using the SQL native `JSON_QUOTE` function, instead of basic string interpolation (`"${address}"`)

## 🚧 Changes

- Use `JSON_QUOTE` to escape input

## 🛠️ Tests

This query should work:

```
query { roles (where: {address: "0xd83261a97fc12d11853bfadae196b832f8aadf79"}) { permissions } }
```

Passing any incomplete address, such as `0x` should not return any result.

### DIFF
--- a/src/graphql/operations/roles.ts
+++ b/src/graphql/operations/roles.ts
@@ -10,14 +10,13 @@ export default async function (parent, args) {
 
   const query = `
     SELECT * FROM spaces
-    WHERE JSON_CONTAINS(LOWER(settings->'$.admins'), LOWER(?))
-      OR JSON_CONTAINS(LOWER(settings->'$.members'), LOWER(?))
-      OR JSON_CONTAINS(LOWER(settings->'$.moderators'), LOWER(?));
+    WHERE JSON_CONTAINS(LOWER(settings->'$.admins'), LOWER(JSON_QUOTE(?)))
+      OR JSON_CONTAINS(LOWER(settings->'$.members'), LOWER(JSON_QUOTE(?)))
+      OR JSON_CONTAINS(LOWER(settings->'$.moderators'), LOWER(JSON_QUOTE(?)));
   `;
-  const params: string[] = [`"${address}"`, `"${address}"`, `"${address}"`];
 
   try {
-    const data = await db.queryAsync(query, params);
+    const data = await db.queryAsync(query, [address, address, address]);
     return data.map((space: any) => {
       const settings = JSON.parse(space.settings);
       const permissions: string[] = [];


src/graphql/operations/roles.ts
```diff
@@ -10,14 +10,13 @@ export default async function (parent, args) {
...
-      OR JSON_CONTAINS(LOWER(settings->'$.moderators'), LOWER(?));
+    WHERE JSON_CONTAINS(LOWER(settings->'$.admins'), LOWER(JSON_QUOTE(?)))
+      OR JSON_CONTAINS(LOWER(settings->'$.members'), LOWER(JSON_QUOTE(?)))
+      OR JSON_CONTAINS(LOWER(settings->'$.moderators'), LOWER(JSON_QUOTE(?)));
   `;
-  const params: string[] = [`"${address}"`, `"${address}"`, `"${address}"`];
 
   try {
-    const data = await db.queryAsync(query, params);
+    const data = await db.queryAsync(query, [address, address, address]);
```
ChaituVR, line 19: Maybe better to check if input is valid address with `getAddress`

ChaituVR (approved):
tAck

src/graphql/operations/roles.ts
wa0x6e: Would be better, but would need to add `@ethersproject/address` as dependency

PR #790, opened 2024-01-13
chore: always run test setup before running test

wa0x6e:
Same setup as sequencer: always run `yarn test:setup` when running `yarn test`


ChaituVR (approved):
utAck

PR #791, opened 2024-01-13
chore: allow node version > 16

wa0x6e:
This PR removes the `engines` property in package.json, to allow more node version.

This change will not affect the node version used on the prod server, which should remain on 16

This service have been tested, and should work on node 18


ChaituVR (approved):
utAck

PR #792, opened 2024-01-15
chore: normalize metrics path

wa0x6e:
Normalizing `graphql` path in metrics, so that `graphql/test/abc`, `graphql?test` etc... are considered as just `graphql`


ChaituVR (approved):
utAck

PR #795, opened 2024-01-15
fix: fix invalid query results

wa0x6e:
Fix https://snapshot-labs.sentry.io/issues/4868101610/?query=is%3Aunresolved&referrer=issue-stream&statsPeriod=30d&stream_index=0

Query is missing the `as count` name


ChaituVR (approved):
utAck

PR #798, opened 2024-01-18
feat: add endpoint to check for deleted spaces

### LINKED ISSUE #796: feat: add new api endpoint to check for deleted spaces
## Goal

Add a new POST `/api/delete-spaces` endpoint, which accept a list of ENS names `['a.eth', 'b.eth']`, and return a JSON list of deleted spaces like `['a.eth']`, or just `[]` if none of the input are deleted spaces

This additional endpoint will avoid having to burden the graphql queries, which does not handle deleted spaces by default.

## Why 

UI does not have a way to check for deleted spaces.
Having that list will allow UI to show filter deleted spaces from the 'Create space' form, and avoid having user completing all the space creation steps, until to get the error "Space deleted, please contact admin" at the end of creation process.

This will lighten the support, which is today getting at least one ticket per week about this error.

See https://github.com/snapshot-labs/snapshot/issues/4516 for UI proposed implementation

wa0x6e:
Fixes #796 and needed by https://github.com/snapshot-labs/snapshot/issues/4516

The GET /spaces/ID endpoint can now returns deleted spaces

Test with

```bash
curl http://localhost:3000/api/spaces/TEST-ID
```

Should return the space settings (if found), with a `deleted` property

### DIFF
--- a/src/api.ts
+++ b/src/api.ts
@@ -42,14 +42,14 @@ router.get('/spaces/:key', async (req, res) => {
 
 router.post('/deleted-spaces', async (req, res) => {
   const { spaces } = req.body;
-  const id = (Array.isArray(spaces) ? spaces : [spaces])
+  const ids = (Array.isArray(spaces) ? spaces : [spaces])
     .map(i => i?.toString())
-    .filter(a => a);
+    .filter(i => i);
 
   try {
     return res.json(
-      id.length > 0
-        ? (await getDeletedSpaces(id)).map((s: { id: string }) => s.id)
+      ids.length > 0
+        ? (await getDeletedSpaces(ids)).map((s: { id: string }) => s.id)
         : []
     );
   } catch (e: any) {
--- a/src/api.ts
+++ b/src/api.ts
@@ -1,6 +1,6 @@
 import express from 'express';
 import { capture } from '@snapshot-labs/snapshot-sentry';
-import { getSpace } from './helpers/spaces';
+import { getSpace, getDeletedSpaces } from './helpers/spaces';
 import { name, version } from '../package.json';
 import { sendError } from './helpers/utils';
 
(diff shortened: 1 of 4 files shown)


src/helpers/spaces.ts
```diff
@@ -197,6 +197,16 @@ export async function getSpace(id: string) {
...
 
+export async function getDeletedSpaces(id: string[]) {
+  const query = `
+    SELECT id
+    FROM spaces
+    WHERE deleted = 1 AND id IN (?)
+    `;
+
+  return await db.queryAsync(query, [id]);
+}
```
ChaituVR, line 208: `id` is array? better to make it `ids`

src/api.ts
```diff
@@ -40,6 +40,24 @@ router.get('/spaces/:key', async (req, res) => {
...
+    return res.json(
+      id.length > 0
+        ? (await getDeletedSpaces(id)).map((s: { id: string }) => s.id)
+        : []
+    );
+  } catch (e: any) {
+    capture(e);
+    return sendError(res, 'server_error', 500);
+  }
+});
```
ChaituVR, line 59: Maybe we can make it a `get` call for now?

ChaituVR (changes requested):

src/api.ts
wa0x6e: GET call with a body ?

ChaituVR: No i mean, a endpoint that can return all deleted spaces, i feel like these spaces will be limited in number, right now there are 600 spaces, but looks like we want to check one space at a time so maybe we can use the route https://hub.snapshot.org/api/spaces/thanku.eth ? and return `deleted` property if spaces is deleted

wa0x6e: The endpoint accept a list spaces, so if user have a lot of ENS in his wallet, we can check all ENS in the same request

wa0x6e:
One another solution is to use the exisring `/api/spaces/SPACE_NAME` to return deleted space, with a `deleted:1` property

ChaituVR:
> One another solution is to use the exisring `/api/spaces/SPACE_NAME` to return deleted space, with a `deleted:1` property

I think that'd be the best solution 👍

wa0x6e:
> > One another solution is to use the exisring `/api/spaces/SPACE_NAME` to return deleted space, with a `deleted:1` property
> 
> I think that'd be the best solution 👍

What’s the current usage of this endpoint? Will that break something if it start returning deleted spaces ?

ChaituVR:
> What’s the current usage of this endpoint? Will that break something if it start returning deleted spaces ?

We mostly use it for dev-related purposes to verify space settings, should not break anything

wa0x6e:
> > What’s the current usage of this endpoint? Will that break something if it start returning deleted spaces ?
> 
> We mostly use it for dev-related purposes to verify space settings, should not break anything

Implemented

ChaituVR (approved):
tAck

PR #800, opened 2024-01-23
fix: only source of proposal flagging should be the DB

wa0x6e:

We use the `flaggedLinks` to mark a proposal as flagged, in addition to the column in the proposals table.

This create inconsistency, as a proposal may be returned as `flagged`, but not when filtering by `flagged` (because the filter does not take flagged links into account)

## 💊 Fixes / Solution

Remove the flaggedLinks handling, and delegate it to sequencer, who will flag the proposal when containing flagged links on creation, thus making the only source of truth for the proposal flag status the `flagged` column in the proposals table.

## 🚧 Changes

- Remove moderation.ts
- Return proposal flagged status directly from the table

## Before merging

Run a SQL query which will retrospectively flag all past proposals containing flaggedLinks


ChaituVR (commented):
weird, if we have this already. what's causing this https://discord.com/channels/955773041898573854/1030772384308932608/1198503030262353950

ChaituVR (approved):
tAck

src/graphql/helpers.ts
```diff
@@ -285,15 +284,6 @@ export function formatUser(user) {
   };
 }
 
-function isFlaggedProposal(proposal) {
-  if (flaggedLinks.length === 0) {
-    return false;
-  }
-
-  const flaggedLinksRegex = new RegExp(flaggedLinks.join('|'), 'i');
-  return flaggedLinksRegex.test(proposal.body) || proposal.flagged;
```
ChaituVR, line 294: This will now not show any notice on proposals with the spam links, unless we flag these proposal,

wa0x6e: Yes, we need to run a SQL query to retrospectively flag all past proposals containing flaggedLinks

PR #805, opened 2024-02-15
fix: allow display of large number values

wa0x6e:
Fixes #805

Fix error when voting period and delay are larger than MAX_INT

The same failing query should now return 

```json
{
  "data": {
    "spaces": [
      {
        "id": "richardfromkfc.eth",
        "name": "APPROVAL PROPOSAL",
        "voting": {
          "period": 864000000000000,
          "delay": 8640000000000000000
        }
      }
    ]
  }
}
```

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -402,8 +402,8 @@ type SpaceFilters {
 }
 
 type SpaceVoting {
-  delay: Int
-  period: Int
+  delay: Float
+  period: Float
   type: String
   quorum: Float
   blind: Boolean


src/graphql/schema.gql
```diff
@@ -402,8 +402,8 @@ type SpaceFilters {
 }
 
 type SpaceVoting {
-  delay: Int
-  period: Int
+  delay: Float
+  period: Float
```
ChaituVR, line 406: `Float` can allow floating point values, don't think it is ideal here

ChaituVR (changes requested):

src/graphql/schema.gql
ChaituVR: Found this 
```javascript
const { GraphQLScalarType } = require('graphql');

const LargeInt = new GraphQLScalarType({
  name: 'LargeInt',
  description: 'Large integer type',
  serialize(value) {
    return value.toString(); // Convert outgoing large integer to string
  },
  parseValue(value) {
    return BigInt(value); // Convert incoming string to large integer
  },
  parseLiteral(ast) {
    if (ast.kind === Kind.STRING) {
      return BigInt(ast.value); // Convert string from query to large integer
    }
    return null;
  },
});
```

ChaituVR:
An important point is that we should never allow this big voting `delay` and `period` 😄 better to limit these to some values, maybe max 30 days delay and 1 year period

src/graphql/schema.gql
wa0x6e: > `Float` can allow floating point values, don't think it is ideal here

It can, but as we have already check in place to ensure these 2 are always integer, graphql will always show these as int, even with the `Float` type.

We can add new check to prevent new large values, but in my mind, editing existing space manually to fix wrong values was never an option.

ChaituVR: > It can, but as we have already check in place to ensure these 2 are always integer, graphql will always show these as int, even with the Float type.

Yes, but `Float` type may confuse graphql API users, where they expect it to float but we always send INT

ChaituVR: > We can add new check to prevent new large values, but in my mind, editing existing space manually to fix wrong values was never an option.

Yea bad option 🙈 but not sure if there is some alternative way

PR #808, opened 2024-02-26
fix: Add new rows to spaces table

ChaituVR:
Run following in your local DB 
```SQL
ALTER TABLE spaces
ADD COLUMN proposal_count INT NOT NULL DEFAULT '0',
ADD COLUMN vote_count INT NOT NULL DEFAULT '0',
ADD COLUMN follower_count INT NOT NULL DEFAULT '0';

CREATE INDEX proposal_count ON spaces (proposal_count);
CREATE INDEX vote_count ON spaces (vote_count);
CREATE INDEX follower_count ON spaces (follower_count);
```


wa0x6e (approved):
utAck

PR #810, opened 2024-02-26
chore: fix `test` command stuck waiting on wrong url

wa0x6e:

The `test` command may be stuck while using `start-server-and-test`, due to it listening on `127.0.0.1` instead of `localhost`. On some env, those 2 urls are the same, but not always (see https://www.npmjs.com/package/start-server-and-test#localhost-vs-0000-vs-127001)

## 💊 Fixes / Solution

`yarn dev` is starting on `localhost`, so `start-server-and-test` should also be explicitly told to listen on `localhost`

## 🚧 Changes

- Update `test` command to listen on the full url instead of just the port

## 🛠️ Tests

- Run `yarn test`
- It should run the test


ChaituVR (approved):
utAck

PR #812, opened 2024-02-26
improve membersURI

Rashmi-278:
Fixes # .

- Get unique Voters for a Space
- Import getUniqueVoters function in EIP4824 File
- Include Unique Voters in the members list for a space

### DIFF
--- a/src/eip4824.ts
+++ b/src/eip4824.ts
@@ -1,5 +1,5 @@
 import express from 'express';
-import { getSpace } from './helpers/spaces';
+import { getSpace, getUniqueVotersForSpace } from './helpers/spaces';
 import db, { sequencerDB } from './helpers/mysql';
 
 const router = express.Router();
@@ -35,21 +35,26 @@ router.get('/:space', async (req, res) => {
 
 router.get('/:space/members', async (req, res) => {
   let space: any = {};
+  let uniqueVoters: any = [];
 
   try {
     space = await getSpace(req.params.space);
+    uniqueVoters = await getUniqueVotersForSpace(req.params.space);
 
     if (!space.verified) return res.status(400).json({ error: 'INVALID' });
   } catch (e) {
     return res.status(404).json({ error: 'NOT_FOUND' });
   }
 
-  const members = [...space.admins, ...space.moderators, ...space.members].map(
-    member => ({
-      type: 'EthereumAddress',
-      id: member
-    })
-  );
+  const members = [
+    ...space.admins,
+    ...space.moderators,
+    ...space.members,
+    ...uniqueVoters
+  ].map(member => ({
+    type: member.type,
+    id: member.id
+  }));
 
   return res.json({
     '@context': context,
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -145,6 +145,23 @@ async function getVotes() {
   return await db.queryAsync(query);
 }
 
+export async function getUniqueVotersForSpace(spaceId) {
+  const query = `
+    SELECT space, GROUP_CONCAT(DISTINCT voter) AS unique_voters
+    FROM votes
+    WHERE space = ?
+    GROUP BY space
+  `;
+  const [result] = await db.queryAsync(query, [spaceId]);
+
+  if (!result) return Promise.reject(new Error('NOT_FOUND'));
+
+  const votersArray = result.unique_voters ? result.unique_voters.split(',') : [];
+
+  return votersArray;
+}
+
+
 async function getFollowers() {
   const query = `
     SELECT space, COUNT(id) as count,


src/helpers/spaces.ts
```diff
@@ -145,6 +145,23 @@ async function getVotes() {
   return await db.queryAsync(query);
 }
 
+export async function getUniqueVotersForSpace(spaceId) {
+  const query = `
+    SELECT space, GROUP_CONCAT(DISTINCT voter) AS unique_voters
+    FROM votes
+    WHERE space = ?
+    GROUP BY space
+  `;
```
bonustrack, line 154: This could be simplified to:
```suggestion
  const query = `SELECT DISTINCT(voter) FROM votes WHERE space = ?`;
```
I think you may want to add a limit here too because in some space there can be millions of voters, which would take too long to resolve.

src/eip4824.ts
```diff
@@ -35,28 +35,54 @@ router.get('/:space', async (req, res) => {
...
+    ...space.members.map(member =>
+      JSON.stringify({ type: 'EthereumAddress', id: member })
+    ),
+    ...uniqueVotersResult.voters.map(voter =>
+      JSON.stringify({ type: 'EthereumAddress', id: voter })
+    )
+  ];
+
+  // Convert the array of JSON strings to a Set to remove duplicates, then back to an array
+  const uniqueMembersJSON = [...new Set(allMembersJSON)];
```
bonustrack, line 70: This is not ideal because you may not always return 100 results, I would rather make a query on the db to enforce that we don't get any duplicated addresses from admin, moderator or authors (members).

src/eip4824.ts
```diff
@@ -35,28 +35,54 @@ router.get('/:space', async (req, res) => {
...
-      type: 'EthereumAddress',
-      id: member
-    })
-  );
-
-  return res.json({
+  // Combine all members and map them to JSON strings for uniqueness check
+  const allMembersJSON = [
+    ...space.admins.map(admin =>
+      JSON.stringify({ type: 'EthereumAddress', id: admin })
```
bonustrack, line 56: You don't need to add the type and id object format as string here, this can be done in the end once you have the final list instead.

bonustrack (changes requested):

src/eip4824.ts
Rashmi-278: I address this by querying the DB with this query

`query = `
    SELECT address
    FROM (
      (SELECT DISTINCT JSON_UNQUOTE(JSON_EXTRACT(settings, '$.admins[*]')) AS address 
       FROM spaces 
       WHERE id = ?)
      UNION
      (SELECT DISTINCT JSON_UNQUOTE(JSON_EXTRACT(settings, '$.members[*]')) AS address 
       FROM spaces 
       WHERE id = ?)
      UNION
      (SELECT DISTINCT JSON_UNQUOTE(JSON_EXTRACT(settings, '$.moderators[*]')) AS address 
       FROM spaces 
       WHERE id = ?)
      UNION
      (SELECT DISTINCT voter AS address
       FROM votes
       WHERE space = ? AND (? IS NULL OR voter > ?)
    ) AS combined_addresses
    ORDER BY address
    LIMIT ?
`;
`
And the last address after the limit reached is considered as cursor

bonustrack (approved):

bonustrack (commented):
I'll review this PR, don't merge yet

.gitignore
```diff
@@ -8,3 +8,5 @@ coverage
 # Remove some common IDE working directories
 .idea
 .vscode
+
+pnpm-lock.yaml
```
bonustrack, line 12: This should be removed, we don't use pnpm on this repository
```suggestion
```

src/eip4824.ts
```diff
@@ -1,9 +1,9 @@
 import express from 'express';
-import { getSpace } from './helpers/spaces';
+import { getCombinedMembersAndVoters, getSpace } from './helpers/spaces';
 import db, { sequencerDB } from './helpers/mysql';
 
 const router = express.Router();
-const context = '<http://www.daostar.org/schemas>';
+const context = ["https://www.snapshot.org", "https://www.daostar.org/schemas"];
```
bonustrack, line 6: We don't use `www` on snapshot.org, also we use single quote, if you run the lint script it should fix that
```suggestion
const context = ['https://snapshot.org', 'https://www.daostar.org/schemas'];
```

bonustrack (changes requested):

src/eip4824.ts
```diff
@@ -34,31 +34,49 @@ router.get('/:space', async (req, res) => {
...
+    console.error(error); 
+
+    if (error.message.includes('database')) {
+      return res.status(500).json({ error: 'DATABASE_ERROR', message: 'Failed to retrieve data from the database.' });
+  } else if (error.message.includes('parameter')) {
+      return res.status(400).json({ error: 'INVALID_PARAMETER', message: 'Invalid or missing parameter.' });
+  } else {
+      return res.status(500).json({ error: 'INTERNAL_SERVER_ERROR', message: 'An unexpected error occurred.' });
+  }
+  }
```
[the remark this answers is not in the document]
Rashmi-278: Yes, done. 
Pardon me.

pnpm-lock.yaml
```diff
@@ -0,0 +1,5756 @@
+lockfileVersion: '6.0'
```
bonustrack, line 1: Can you remove this file?

Rashmi-278: Done.

src/helpers/spaces.ts
```diff
@@ -145,6 +145,57 @@ async function getVotes() {
...
+       WHERE id = ?)
+      UNION
+      (SELECT DISTINCT voter AS address
+       FROM votes
+       WHERE space = ?)
+    ) AS combined_addresses
+    WHERE (? IS NULL OR address > ?)
+    ORDER BY address
+    LIMIT ?
+  `;
```
ChaituVR, line 176: This is actually taking 22 seconds for spaces like stgdao.eth

src/helpers/spaces.ts
```diff
@@ -145,6 +145,57 @@ async function getVotes() {
...
+       FROM spaces 
+       WHERE id = ?)
+      UNION
+      (SELECT DISTINCT JSON_UNQUOTE(JSON_EXTRACT(settings, '$.members[*]')) AS address 
+       FROM spaces 
+       WHERE id = ?)
+      UNION
+      (SELECT DISTINCT voter AS address
+       FROM votes
+       WHERE space = ?)
```
ChaituVR, line 171: Should limit to `500` maybe? 
```suggestion
      (SELECT DISTINCT voter AS address
       FROM votes
       WHERE space = ? LIMIT 500)
```

ChaituVR (changes requested):

src/helpers/spaces.ts
ChaituVR, line 171: should limit to 500 voters

src/eip4824.ts
```diff
@@ -34,29 +34,61 @@ router.get('/:space', async (req, res) => {
...
-    return res.status(404).json({ error: 'NOT_FOUND' });
-  }
+    // Includes space.members, space.admins, space.moderators and voters
+    const combinedMembersResult = await getCombinedMembersAndVoters(
+      spaceId,
+      cursor,
+      pageSize
+    );
 
-  const members = [...space.admins, ...space.moderators, ...space.members].map(
```
ChaituVR, line 47: here we already have `admins` , `moderators` and `members` , is there any reason to get them again in `getCombinedMembersAndVoters` ? If there is no big reason better to read them from here to reduce some load on DB

ChaituVR:
Apart from above comments, everything looks to be working nicely 👍

src/eip4824.ts
Rashmi-278: `getCombinedMembersAndVoters` Function is designed to get all the members of the DAO, including Admins, Moderators, Members and Voters. 
 
 To get a unique, consolidated list of all member groups, including pagination we are enforcing this on DB query.
 
 In my previous approach, this was handled at the API level.
  Here's a reference to my old approach,
 https://github.com/snapshot-labs/snapshot-hub/pull/812/commits/3e5bf86cadec031b9cc6cad7240676a05a5a748d

src/helpers/spaces.ts
Rashmi-278: Noted. 22 secs for a DAO with 912K members

src/helpers/spaces.ts
Rashmi-278: I have placed the limit to the Outer Query, using the param, pageSize.
Placing a limit on the inner query would mess up the pagination, wouldn't it?

ChaituVR: Yes but this is taking a lot of time, especially for spaces with a lot of votes, we don't need to read all the distinct voters here, we just need 500 of them, 

you can remove pagination from the outer query

Also for `admins`, `moderators` and `members` -> all this data is already available on `space` object, we don't need to read them from database again,

src/eip4824.ts
ChaituVR: > To get a unique, consolidated list of all member groups, including pagination we are enforcing this on DB query.

In that case we can exclude `admins`, `moderators` and `members` from voters within the SQL query,

Rashmi-278: Changes updated.

src/helpers/spaces.ts
```diff
@@ -145,6 +145,51 @@ async function getVotes() {
...
+    SELECT DISTINCT voter AS address
+    FROM votes
+    WHERE space_id = ? AND voter NOT IN (${placeholders})
+  `);
+  params.push(spaceId, ...exclusionList);
+
+  const cursorClause = cursor ? ' AND address > ?' : '';
+  const query = `
+    SELECT address
+    FROM (${subqueries.join(' UNION ')})
```
ChaituVR, line 167: I get error `Every derived table must have its own alias`

ChaituVR (changes requested):

src/helpers/spaces.ts
ChaituVR: we always get one item in `subqueries` right? maybe we can use a string instead of array

ChaituVR: It worked after changing it to string and changing above line to:
```suggestion
     FROM (${subqueries}) as subqueries
```

src/helpers/spaces.ts
```diff
@@ -145,6 +145,51 @@ async function getVotes() {
...
+
+  // Excluding known addresses fetched during the Space verification
+  const exclusionList = [...knownAdmins, ...knownModerators, ...knownMembers];
+  let placeholders = exclusionList.map(() => '?').join(', ');
+
+  // Other roles are already known and fetched at the app level while Space Verification
+  subqueries.push(`
+    SELECT DISTINCT voter AS address
+    FROM votes
+    WHERE space_id = ? AND voter NOT IN (${placeholders})
```
ChaituVR, line 160: There is no such field as `space_id`, it should be `space` 
```suggestion
    WHERE space = ? AND voter NOT IN (${placeholders})
```

src/eip4824.ts
```diff
@@ -34,31 +34,68 @@ router.get('/:space', async (req, res) => {
 });
 
 router.get('/:space/members', async (req, res) => {
-  let space: any = {};
-
   try {
-    space = await getSpace(req.params.space);
+    const spaceId = req.params.space;
+    const cursor = req.query.cursor || null;
```
ChaituVR, line 39: When i try http://localhost:3000/api/eip4824/ens.eth/members?cursor=1 i get `Error: NOT_FOUND` error

src/eip4824.ts
```diff
@@ -34,31 +34,68 @@ router.get('/:space', async (req, res) => {
...
+      space.moderators || [],
+      space.members || []
+    );
+
+    const members = [
+      ...space.admins.map(admin => ({ type: 'EthereumAddress', id: admin })),
+      ...space.moderators.map(moderator => ({ type: 'EthereumAddress', id: moderator })),
+      ...space.members.map(member => ({ type: 'EthereumAddress', id: member })),
+      ...combinedMembersResult.members.map(voter => ({ type: 'EthereumAddress', id: voter }))
+    ];
```
ChaituVR, line 64: I feel that irrespective of `cursor` input, this will always return admins, moderators, and members on top

src/eip4824.ts
Rashmi-278: Yes, It is expected behavior, cursor is supposed is to be an Ethereum wallet address, if that wallet address is not of a member or an invalid address it returns NOT_FOUND error.

src/eip4824.ts
ChaituVR: @Rashmi-278 Could you please check this?

ChaituVR: using cursor we should exclude the members

src/helpers/spaces.ts
Rashmi-278: We made an improvement on this, now the with cursor filter,  if cursor is present, admins, moderators, and members are excluded from the distinct voter list. Also regardless we are not querying admins, moderators and members.

Query now only queries the Votes Table hence eliminating the need for subqueries.

src/eip4824.ts
```diff
@@ -34,31 +34,84 @@ router.get('/:space', async (req, res) => {
...
+    if (cursor) {
+      const combinedMembersResult = await getCombinedMembersAndVoters(
+        spaceId,
+        cursor,
+        pageSize,
+        [], 
+        [], 
+        []  
+      );
+      members = combinedMembersResult.members.map(voter => ({ type: 'EthereumAddress', id: voter }));
```
ChaituVR, line 62: lint seems off, could you please run `yarn lint`

src/eip4824.ts
Rashmi-278: This improvement is added.

src/helpers/spaces.ts
ChaituVR: how to use cursor now? you have an example request? i tried few ways but couldn't get it working

Rashmi-278: A request should be something like` /ens.eth/members?cursor=983434323` ( Cursor Example )

Could you share what message do you get?
Could we get on a call and debug this? As I don't have a way to see whats wrong, its taking longer to close this issue

src/eip4824.ts
Rashmi-278: Lint done

src/helpers/spaces.ts
```diff
@@ -219,6 +219,53 @@ async function getVotes(): Promise<Record<string, { votesCount7d: number }>> {
...
+) {
+  const params: (string | number)[] = [spaceId];
+  const exclusionList = [...knownAdmins, ...knownModerators, ...knownMembers];
+
+  // Other roles are already known and fetched at the app level while Space Verification
+  // Building the exclusion clause only if the exclusion list is not empty
+  let exclusionClause = '';
+  if (exclusionList.length > 0) {
+    const placeholders = exclusionList.map(() => '?').join(', ');
+    exclusionClause = `AND voter NOT IN (${placeholders})`;
```
bonustrack, line 238: Can't this be used to inject arbitrary SQL? I think `placeholders` should be added as params so it get parsed.

Rashmi-278: I don't think that would be of concern here because 
1. Placeholders values are not user input values
2. Placeholder values are either empty lists or space.admins, space.moderators, space.members
3. `params.push(...exclusionList)`  This pushes the actual exclusion list values (the voters you want to exclude) into the params array, which is passed to the query execution.
4. ` const results =await db.queryAsync(query, params);`  The database library takes care of escaping the values in params, preventing SQL injection.

Issue #444, opened 2022-09-01
Vote query on GraphQL is not working


ChaituVR:
Reported by @hjjlxm on discord

https://hub.snapshot.org/graphql?operationName=Vote&query=query%20Vote%20%7B%0A%20%20vote%20(%0A%20%20%20%20id%3A%20%220x5e2bbca461950efae33244fbf7707b8f907780d7521f331227956692adff9862%22%0A%20%20)%20%7B%0A%20%20%20%20id%0A%20%20%20%20voter%0A%20%20%20%20created%0A%20%20%20%20proposal%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%20%20choice%0A%20%20%20%20space%20%7B%0A%20%20%20%20%20%20id%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D

Irrespective of what we give as input `id` it always fail with error:
```json
{
  "errors": [
    {
      "message": "Unexpected error value: \"request failed\"",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "vote"
      ]
    }
  ],
  "data": {
    "vote": null
  }
}
```

ChaituVR:
Also should be possible to find a vote with IPFS instead of just id (UI just display IPFS)

hjjlxm:
> Also should be possible to find a vote with IPFS instead of just id (UI just display IPFS)

Thanks for raising this. Vote id is more abstract and cannot be accessed through Receipt (only ipfs id). Therefore, using ipfs id instead as a condition for vote query is more intuitive

ChaituVR:
Update: can't reproduce this in my local, issue must be with production DB, 

Error:
```
sqlMessage: \"Illegal mix of collations (utf8mb4_unicode_ci,IMPLICIT) and (utf8mb4_0900_ai_ci,IMPLICIT) for operation '='\",
```

wa0x6e:
I can confirm that this is a DB issue, by running the underlying Mysql query directly on the db give the same error.

wa0x6e:
Running the following SQL command 

```
SELECT TABLE_SCHEMA 
    , TABLE_NAME 
    , COLUMN_NAME 
    , COLLATION_NAME 
FROM INFORMATION_SCHEMA.COLUMNS
WHERE TABLE_schema = 'defaultdb';
```

We can see that the `votes` table is not using the same collation as all the other tables.

[...]

PR #815, opened 2024-03-01
fix: vote query not working

ChaituVR:
fixes https://github.com/snapshot-labs/snapshot-hub/issues/444

### Solution
To change collation but that is not happening so calling the `votes` query will fix this for now

### How to test
- Make sure your `votes` table and `spaces` tables have different collations
- try the following query
```Graphql
query Vote {
  vote (
    id: "0xf6c08b692d4bd1fd4c091431895309d38f3d6d7770c075c14b63fca27662b073"
  ) {
    id
    space {
      id
    }
  }
}
```
- Before fix it should return error `Unexpected error value: \"request failed\"` 
- After fix it should return vote response

### DIFF
--- a/src/graphql/operations/vote.ts
+++ b/src/graphql/operations/vote.ts
@@ -1,49 +1,19 @@
-import graphqlFields from 'graphql-fields';
-import db from '../../helpers/mysql';
-import { formatProposal, formatVote } from '../helpers';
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

wa0x6e (approved):
tAck

PR #817, opened 2024-03-06
fix: Include strategy networks and multichain networks in network count

ChaituVR:
Right now we are counting main space network, this change will add
- Strategy network count
- Networks used inside multichain strategy

Before change:
image

After change: 
image


wa0x6e (approved):
tAck

PR #823, opened 2024-03-14
feat: Add boost fields to space query

### LINKED ISSUE #822: Add Boost settings to spaces query
```[tasklist]
### Tasks
- [ ] https://github.com/snapshot-labs/snapshot/pull/4616
- [ ] https://github.com/snapshot-labs/snapshot.js/pull/976
- [x] Add boost settings to graphql schema
- [x] By default it should return ```{enabled: true, bribeEnabled: false}```
```

ChaituVR:
Fixes #822 

Add boost fields to space/spaces query so UI can retrieve them

### How to test:
- Try a query like 
```graphQL
{
  space(id:"thanku.eth") {
    id
    boost {
      enabled
      bribeEnabled
    }
  }
}
```
- It should return default values
- Now try changing the settings of your space directly on DB (or use UI https://github.com/snapshot-labs/snapshot/pull/4616 to update your DB settings) 
-  On you DB, try adding:
```JSON
"boost":{"enabled":true,"bribeEnabled":true}
```
- Now try the above query again, it should return updated values

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -83,6 +83,8 @@ export function formatSpace({
   space.voting.hideAbstain = space.voting.hideAbstain || false;
   space.voteValidation = space.voteValidation || { name: 'any', params: {} };
   space.delegationPortal = space.delegationPortal || null;
+  // disable boost by default until it's enabled
+  space.boost = space.boost || { disabled: true, bribeEnabled: false };
   space.strategies = space.strategies?.map(strategy => ({
     ...strategy,
     // By default return space network if strategy network is not defined
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -83,8 +83,7 @@ export function formatSpace({
   space.voting.hideAbstain = space.voting.hideAbstain || false;
   space.voteValidation = space.voteValidation || { name: 'any', params: {} };
   space.delegationPortal = space.delegationPortal || null;
-  // disable boost by default until it's enabled
-  space.boost = space.boost || { disabled: true, bribeEnabled: false };
+  space.boost = space.boost || { enabled: true, bribeEnabled: false };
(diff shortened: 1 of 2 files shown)


wa0x6e:
Would it be more logical to have the field be `enabled` instead of `disabled`, so it's more consistent with `bribeEnabled` ?

src/graphql/helpers.ts
```diff
@@ -83,6 +83,8 @@ export function formatSpace({
   space.voting.hideAbstain = space.voting.hideAbstain || false;
   space.voteValidation = space.voteValidation || { name: 'any', params: {} };
   space.delegationPortal = space.delegationPortal || null;
+  // disable boost by default until it's enabled
+  space.boost = space.boost || { disabled: true, bribeEnabled: false };
```
wa0x6e, line 87: Would it be more logical to have the field be `enabled` instead of `disabled`, so it's more consistent with `bribeEnabled` ?

ChaituVR: So i talked to @samuveth about it, https://github.com/snapshot-labs/snapshot.js/pull/975#discussion_r1524823252 The future plan is to enable it to everyone by default and only disable it to who want to disable it

samuveth:
Can you make the default values both `false`

samuveth:
As I mentioned in DMs I will handle disabling with whitelist now, so it can be `false` here by default

src/graphql/helpers.ts
samuveth: But if we have default values I can also call it `enabled`, and it could default to `true`

wa0x6e: > But if we have default values I can also call it `enabled`, and it could default to `true`

I prefer this

samuveth (approved):
utACK

wa0x6e (approved):
utAck

PR #827, opened 2024-03-19
Add Quorum type on proposal and space

ChaituVR:
Related https://github.com/snapshot-labs/snapshot/issues/4600

Returns `quorumType` on proposal and space

### How to test:
- Change `quorumType` in your space settings
- Create proposals with different `quorumType`
Try running a query like:
```
{
  spaces {
   id
    voting{
      quorumType
      quorum
    }
  }
}

{
  proposals {
    id
    quorum
    quorumType
  }
}
```
- Should return `default` by default and `optimistic` if changed

- [ ] On Hub DB run:
```SQL
ALTER TABLE proposals
ADD COLUMN quorum_type VARCHAR(24) DEFAULT '' AFTER quorum;
```

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -407,6 +407,7 @@ type SpaceVoting {
   period: Int
   type: String
   quorum: Float
+  quorumType: String
   blind: Boolean
   hideAbstain: Boolean
   privacy: String
@@ -433,6 +434,7 @@ type Proposal {
   start: Int!
   end: Int!
   quorum: Float!
+  quorumType: String!
   privacy: String
   snapshot: String
   state: String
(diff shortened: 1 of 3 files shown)


src/graphql/schema.gql
```diff
@@ -433,6 +434,7 @@ type Proposal {
   start: Int!
   end: Int!
   quorum: Float!
+  quorumType: String!
```
wa0x6e, line 437: Why is this one using `String!`, and the one from SpaceVoting using just `String` ?

src/graphql/schema.gql
```diff
@@ -407,6 +407,7 @@ type SpaceVoting {
   period: Int
   type: String
   quorum: Float
+  quorumType: String
```
ChaituVR, line 410:
```suggestion
  quorumType: String!
```

src/graphql/schema.gql
ChaituVR: Updated, Thank you!

wa0x6e (approved):
utAck

PR #831, opened 2024-03-27
fix: Reduce wait time to two mins while loading spaces

ChaituVR:
Reducing wait time to 2 mins from 6 mins, as 6 mins is quite high and it feels like something is down

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -206,6 +206,6 @@ export default async function run() {
     capture(e);
     log.error(`[spaces] failed to load spaces, ${JSON.stringify(e)}`);
   }
-  await snapshot.utils.sleep(360e3);
+  await snapshot.utils.sleep(120e3);
   run();
 }


wa0x6e:
Why would it feels like something is down ? This function is run immediately on server start

ChaituVR:
> Why would it feels like something is down ? This function is run immediately on server start

No not on server start, but try going to https://testnet.snapshot.org/#/ and follow a space 😄 it takes very long to see correct followers, i think these functions `loadSpaces` `loadSpacesMetrics` take long time on prod anyway, I feel waiting for 6 mins doesn't make anything better.

wa0x6e:
I see now. Theses counters will be faster with https://github.com/orgs/snapshot-labs/projects/12/views/1?pane=issue&itemId=55693630 later, and this `run()` should only updates space scores used for sorting

wa0x6e (approved):
utAck

PR #834, opened 2024-04-05
Fix: vote validation default value

ChaituVR:
This one is a temporary fix for https://github.com/snapshot-labs/snapshot-sequencer/pull/325/files to return a default value when `validation` is `{}`

ideally, we will save 
```js
{
      name: 'any',
      params: {}
}
```
or `null` to this field


wa0x6e (approved):
utAck

wa0x6e:
Can not save null to this field, as the schema does not permit null values

PR #836, opened 2024-04-08
fix: throw error on missing required query argument

### LINKED ISSUE #835: Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your M...
Sentry Issue: [SNAPSHOT-HUB-2E](https://snapshot-labs.sentry.io/issues/5049433005/?referrer=github_integration)

```
Error: ER_PARSE_ERROR: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '? LIMIT 1' at line 1
  File "/home/node/app/dist/src/graphql/operations/user.js", line 21, in Object.default_1 [as user]
    (0, snapshot_sentry_1.capture)(e, { args });
```

wa0x6e:
Fixes #835 
Fixes [SNAPSHOT-HUB-2E](https://snapshot-labs.sentry.io/issues/5049433005/?referrer=github_integration)

- Enforce presence of required argument in the gql schema, so that error is thrown when missing, else it will continue and fail with an invalid SQL query error.

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -3,7 +3,7 @@ scalar Any
 # Queries
 
 type Query {
-  space(id: String): Space
+  space(id: String!): Space
 
   spaces(
     first: Int! = 20
@@ -15,7 +15,7 @@ type Query {
 
   ranking(first: Int! = 20, skip: Int! = 0, where: RankingWhere): RankingObject
 
-  proposal(id: String): Proposal
+  proposal(id: String!): Proposal
 
   proposals(
     first: Int! = 20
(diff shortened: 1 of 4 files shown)


src/graphql/schema.gql
```diff
@@ -3,7 +3,7 @@ scalar Any
 # Queries
 
 type Query {
-  space(id: String): Space
+  space(id: String!): Space
```
ChaituVR, line 6: I think with this we can remove this line https://github.com/snapshot-labs/snapshot-hub/blob/master /src/graphql/operations/space.ts#L6 and for vote https://github.com/snapshot-labs/snapshot-hub/blob/master /src/graphql/operations/vote.ts#L7

ChaituVR (changes requested):

ChaituVR (approved):
tAck

PR #839, opened 2024-04-16
feat: add endpoint for leaderboards

wa0x6e:
Following https://github.com/snapshot-labs/snapshot-sequencer/pull/127

## Add endpoint to return data from `leaderboard` table

Test with the query 

```graphql
{
  leaderboards(where: { vote_count_gte: 10, space_in: ["ens.eth"] }) {
    user 
    space 
    votesCount 
    proposalsCount
    lastVote
  }
}
```

This require that you have run the refresh_leaderboard script from sequencer to populate the table first. `user` and `space` can return the full User and Space object.

Leaderboard API is written in order to match the format used in https://api-1.snapshotx.xyz/

## Add `proposal_count` and `vote_count` to `User` object

Thanks to the new leaderboard table, we now have access to the total number of votes and proposals created by an user.

```graphql
{
  user(id: "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7") {
    id 
    votesCount
    proposalsCount
  }
}
```

Will return 

```
{
  "data": {
    "user": {
      "id": "0xeF8305E140ac520225DAf050e2f71d5fBcC543e7",
      "votesCount": 224,
      "proposalsCount": 217
    }
  }
}
```

Those 2 new properties have been added to match the `User` object returned by the SX api

## Note 

Return properties name are `proposalsCount` and `votesCount` for consistency with existing hub API, and not matching the `proposal_count` and `vote_count` name on sx


src/graphql/operations/leaderboards.ts
```diff
@@ -0,0 +1,58 @@
...
+  let orderDirection = (args.orderDirection || 'desc').toUpperCase();
+  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+
+  const query = `
+    SELECT l.*,
+      spaces.settings,
+      spaces.flagged as spaceFlagged,
+      spaces.verified as spaceVerified,
+      spaces.turbo as spaceTurbo,
+      spaces.hibernated as spaceHibernated,
```
bonustrack, line 32: Why you need this here?

wa0x6e: Those are properties of the `Space` object.

First version of this PR was returning result like 

```
"space": "ens.eth",
"user": "0xcEF0b1BbEC22eDC3D07Eac2F5f1053359A96A5e2",
"vote_count": 38,
"proposal_count": 0
```

Now, it's returning 

```
{
        "user": {
          "id": "0xcEF0b1BbEC22eDC3D07Eac2F5f1053359A96A5e2",
          "ipfs": "bafkreiai7c3canwhltsxjtc4ba2fli4mlrn7plmo4nl27cp27cugzsxl44",
          "created": 1671060995,
          "name": "",
          "about": "",
          "avatar": "ipfs://bafkreiguy3xfr246zcucsnk7lvneck74jb5rp3nvcqapfbveutwwimspo4"
        },
        "space": {
          "id": "ens.eth",
          "activeProposals": 0,
          "verified": true
        },
        "vote_count": 38,
        "proposal_count": 0
      },
```

with `space` and `user` returning object instead of just the ID, to match the API on sx

src/helpers/schema.sql
```diff
@@ -157,3 +157,14 @@ CREATE TABLE statements (
   INDEX created (created),
   INDEX updated (updated)
 );
+
+CREATE TABLE leaderboard (
```
ChaituVR, line 161: also update last_vote

src/graphql/schema.gql
```diff
@@ -566,3 +603,10 @@ type Vp {
   vp_by_strategy: [Float]
   vp_state: String
 }
+
+type Leaderboard {
+  space: Space
+  user: User
+  proposalsCount: Int
+  votesCount: Int
+}
```
ChaituVR, line 612: lastVote is missing here also. why do we wanted to store this?

src/graphql/operations/leaderboards.ts
ChaituVR: I think `space` in leaderboard could be just string, else we would make it slow like other queries. in most of the times we would query leaderboard of single space right? 

same with `user` but could be helpful, is it possible to compare times with and without these two tables here?

wa0x6e: `space` and `user` now returned as simple string, avoiding all `JOIN` in the sql query

wa0x6e:
From my tests, the query is taking ~3s, due to the sorting with `votesCount DESC, proposalsCount DESC`

- Removing all sorting drop the query to 400ms
- Using only one field to sort instead of 2 drop the query to 800ms

Leaderboard is a ranking of top contributors, should be obvious to sort by proposals count if 2 users have the same number of votes, else order will be random

src/graphql/operations/user.ts
```diff
@@ -5,7 +5,16 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
...
+  const query = `
+    SELECT
+      u.*,
+      SUM(l.vote_count) as votesCount,
+      SUM(l.proposal_count) as proposalsCount,
+      MAX(l.last_vote) as lastVote
+    FROM users u
+    LEFT JOIN leaderboard l ON BINARY l.user = BINARY u.id
+    WHERE id = ?
+    LIMIT 1`;
```
ChaituVR, line 17: Looks like this take more than 30 secs

ChaituVR: As you said Binary casting is causing this, we should do something like this
```suggestion
  const query = `
    SELECT
      u.*,
      SUM(l.vote_count) as votesCount,
      SUM(l.proposal_count) as proposalsCount,
      MAX(l.last_vote) as lastVote
    FROM users u
    LEFT JOIN leaderboard l ON l.user = ?
    WHERE id = ?
    LIMIT 1`;
```

ChaituVR: Actually, it takes very less we can just use `l.user = BINARY u.id` instead of `BINARY l.user = BINARY u.id`

src/graphql/operations/users.ts
```diff
@@ -24,16 +24,21 @@ export default async function (parent, args) {
...
+    SELECT
+      u.*,
+      SUM(l.vote_count) as votesCount,
+      SUM(l.proposal_count) as proposalsCount,
+      MAX(l.last_vote) as lastVote
+    FROM users u
+    INNER JOIN leaderboard l ON BINARY l.user = BINARY u.id
     WHERE 1=1 ${queryStr}
+    GROUP BY u.id
     ORDER BY ${orderBy} ${orderDirection} LIMIT ?, ?
```
ChaituVR, line 37: Same here 

```suggestion
    SELECT
      u.*,
      SUM(l.vote_count) as votesCount,
      SUM(l.proposal_count) as proposalsCount,
      MAX(l.last_vote) as lastVote
    FROM users u
    INNER JOIN leaderboard l ON l.user = BINARY u.id
    WHERE 1=1 ${queryStr}
    GROUP BY u.id
    ORDER BY ${orderBy} ${orderDirection} LIMIT ?, ?
```

src/graphql/operations/user.ts
ChaituVR:
```suggestion
  const query = `
    SELECT
      u.*,
      SUM(l.vote_count) as votesCount,
      SUM(l.proposal_count) as proposalsCount,
      MAX(l.last_vote) as lastVote
    FROM users u
    LEFT JOIN leaderboard l ON l.user = BINARY u.id
    WHERE id = ?
    LIMIT 1`;
```

ChaituVR (changes requested):

wa0x6e:
True that it takes less time, using BINARY casting on only one column.

Still, I just think this PR should be merged only after we resolve the COLLATION issue, and just remove all `BINARY` casting (and keep the JOIN)

wa0x6e:
Refactored to remove all `BINARY` casting, query seems to take 200ms now, as fast as the one without JOIN on the leaderboard

ChaituVR (approved):
tAck

PR #843, opened 2024-04-28
fix: fix error when variable is null

### LINKED ISSUE #841: TypeError: Cannot read properties of null (reading 'toLowerCase')
Sentry Issue: [SNAPSHOT-HUB-2M](https://snapshot-labs.sentry.io/issues/5259562175/?referrer=github_integration)

```
TypeError: Cannot read properties of null (reading 'toLowerCase')
  File "/home/node/app/dist/src/graphql/operations/ranking.js", line 20, in Object.default_1 [as ranking]
    const searchStr = search.toLowerCase();
...
(5 additional frame(s) were not displayed)
```

wa0x6e:
Fixes #841 .

Fix an error when variable is `null` instead of `string` type

### DIFF
--- a/src/graphql/operations/ranking.ts
+++ b/src/graphql/operations/ranking.ts
@@ -26,8 +26,8 @@ export default async function (_parent, args, context, info) {
       strategy = '',
       plugin = ''
     } = where;
-    const searchStr = search.toLowerCase();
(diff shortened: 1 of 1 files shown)


bonustrack (approved):
utACK

PR #844, opened 2024-04-28
feat: add `proposalsCount` and `votesCount` to users

wa0x6e:

We don't have a `proposal_count` and `vote_count` for the `User` object,

## 💊 Fixes / Solution

Add those missing properties, to be inline with the API from sx 

Screenshot 2024-04-28 at 22 53 13

## 🚧 Changes

- Return a `proposal_count` and `vote_count` properties for the User object (computed from the leaderboard table)
- Add filters to `users`, to filter them by proposals and votes count

## 🛠️ Tests

```graphql
{
  user(id: "0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3") {
   	id  
    proposal_count
    vote_count
  }
}

// Should return 

{
  "data": {
    "user": {
      "id": "0x91FD2c8d24767db4Ece7069AA27832ffaf8590f3",
      "proposal_count": 9,
      "vote_count": 13
    }
  }
}
```

```graphql
{
  users(where: { proposal_count_gte: 10, vote_count_gte: 1 }) {
   	id  
    proposal_count
    vote_count
  }
}

// Should return 

{
  "data": {
    "users": [
      {
        "id": "0xd333Bc5c9670C9cEb18f9A2CF02C6E86807a8227",
        "proposal_count": 12,
        "vote_count": 5
      },
      {
        "id": "0xa561e70cd46B6445D98Ffa26e0288C9E49C70B1a",
        "proposal_count": 10,
        "vote_count": 8
      },
      {
        "id": "0x9f74662aD05840Ba35d111930501c617920dD68e",
        "proposal_count": 11,
        "vote_count": 7
      }
    ]
  }
}
```

## TODO 

- [ ] Waiting for #839 to be merged first

### DIFF
--- a/src/graphql/operations/users.ts
+++ b/src/graphql/operations/users.ts
@@ -8,36 +8,52 @@ export default async function (parent, args) {
 
   checkLimits(args, 'users');
 
-  const fields = {
-    id: 'string',
-    ipfs: 'string',
-    created: 'number'
-  };
+  const fields = { id: 'string', ipfs: 'string', created: 'number' };
+  const leaderboardFields = { vote_count: 'number', proposal_count: 'number' };
+
   const whereQuery = buildWhereQuery(fields, 'u', where);
-  const queryStr = whereQuery.query;
-  const params: any[] = whereQuery.params;
-
-  let orderBy = args.orderBy || 'created';
-  let orderDirection = args.orderDirection || 'desc';
-  if (!['created'].includes(orderBy)) orderBy = 'created';
-  orderBy = `u.${orderBy}`;
-  orderDirection = orderDirection.toUpperCase();
-  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
-
-  let users: any[] = [];
+  const leaderboardWhereQuery = buildWhereQuery(leaderboardFields, 'l', where);
+  const queryStr = [whereQuery.query, leaderboardWhereQuery.query].join(' ');
+  const params = whereQuery.params.concat(
+    leaderboardWhereQuery.params,
+    skip,
+    first
+  );
 
   const query = `
-    SELECT u.* FROM users u
+    SELECT
+      u.*,
+      SUM(l.vote_count) AS vote_count,
+      SUM(l.proposal_count) AS proposal_count
+    FROM users u
+    JOIN leaderboard l ON BINARY l.user = BINARY u.id
     WHERE 1=1 ${queryStr}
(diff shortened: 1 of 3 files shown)


ChaituVR (changes requested):
We are still waiting to run script for leaderboard no? this may return wrong counts if we merge

src/graphql/schema.gql
```diff
@@ -521,6 +533,8 @@ type User {
   name: String
   about: String
   avatar: String
+  vote_count: Int
+  proposal_count: Int
```
ChaituVR, line 537: Maybe we could add last_vote also?

ChaituVR (changes requested):

src/graphql/operations/users.ts
```diff
@@ -8,36 +8,52 @@ export default async function (parent, args) {
...
-  const queryStr = whereQuery.query;
-  const params: any[] = whereQuery.params;
-
-  let orderBy = args.orderBy || 'created';
-  let orderDirection = args.orderDirection || 'desc';
-  if (!['created'].includes(orderBy)) orderBy = 'created';
-  orderBy = `u.${orderBy}`;
-  orderDirection = orderDirection.toUpperCase();
-  if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+  const fields = { id: 'string', ipfs: 'string', created: 'number' };
```
ChaituVR, line 11: There is no way to filter with space?

wa0x6e: a `space` filter does not make sense on this `User` object. Maybe you're talking about Leaderboard ?

src/graphql/schema.gql
wa0x6e: We could, but it's not part of SX api, and will not be used anywhere (for now)

src/graphql/operations/users.ts
ChaituVR: oh yes I mean leaderboard 😄 why do we want to filter with vote_count and proposal_count but not space? 🤔

src/graphql/schema.gql
ChaituVR: I think better to add it anyway to know the last voted timestamp of a user, but your decision, not too important

src/graphql/operations/users.ts
wa0x6e: Filter by vote_count to for example fetch only user who have vote_count < X

Filterting by space will seem weird, like "I want to filter users with space `abc.eth`". What would that filter mean ?

wa0x6e:
Duplicate with #839

PR #847, opened 2024-05-02
feat: add support for `network` field in follows

wa0x6e:
Fixes https://github.com/snapshot-labs/snapshot-sequencer/issues/341

This PR add support for the new `network` field added to follows in https://github.com/snapshot-labs/snapshot-sequencer/pull/342

- Add `network` and `network_in` filter to follows
- Return the `network` field in follows

As we only have Space data for followed spaced of type `s`, all the spaces from other network will only return an `id`, with all other fields blank.

## Tests 

```graphql
{
  follows(where: {network: "s"}, first: 10) {
   id
    space {
      id 
      name
      about
      turbo
      flagged
      verified
    }
    network
  }
}
```

### DIFF
--- a/src/graphql/operations/follows.ts
+++ b/src/graphql/operations/follows.ts
@@ -13,11 +13,12 @@ export default async function (parent, args) {
     ipfs: 'string',
     follower: 'string',
     space: 'string',
+    network: 'string',
     created: 'number'
   };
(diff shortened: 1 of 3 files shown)


ChaituVR (approved):
tAck

PR #854, opened 2024-05-28
fix: Remove un-used id and id_in from RankingWhere

ChaituVR:
These inputs were never handled in rankings query


wa0x6e:
Need to fix this one first (https://github.com/snapshot-labs/sx-monorepo/issues/366) else it'll break all queries

wa0x6e (approved):
utAck

PR #857, opened 2024-05-31
feat: add `controller` to space filters

wa0x6e:
This PR add `controller` to `Space` filters, we we could get all spaces managed by a given address, like

```graphql
query Spaces {
  spaces(
    first: 20,
    skip: 0,
    orderBy: "created",
    orderDirection: desc,
    where: { controller: "0x24F15402C6Bb870554489b2fd2049A85d75B982f" }
  ) {
    id
  }
}
```

This avoid having the UI to lookup the address' domain, then send the space id with `id_in` when fetching all spaces managed by a given address (like in the "My Spaces" page)

This also enable API parity with SX api, which is already using the graphql API to show all spaces for a given controller, instead of resolving them UI side.

This is the first step of our transition into having a hardcoded `controller` property, on the space object

### TODO 

- [x] Merge https://github.com/snapshot-labs/stamp/pull/231


src/graphql/helpers.ts
```diff
@@ -375,3 +383,35 @@ export function formatSubscription(subscription) {
...
+    result: string[];
+    error?: {
+      code: number;
+      message: string;
+      data: any;
+    };
+  };
+
+  try {
+    const response = await fetch(process.env.STAMP_URL, {
```
ChaituVR, line 398: Could we add a default stamp url if env doesn't exist?

src/graphql/helpers.ts
```diff
@@ -375,3 +383,35 @@ export function formatSubscription(subscription) {
...
+      method: 'POST',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json'
+      },
+      body: JSON.stringify({
+        method: 'lookup_domains',
```
ChaituVR, line 405: But we have other conditions to get space controller 🤔  Refer to this function https://github.com/snapshot-labs/snapshot.js/blob/master/src/utils.ts#L590-L607

ChaituVR (changes requested):

src/graphql/helpers.ts
wa0x6e: This is true, but this is not the method used by the UI to fetch all spaces controlled by an user.

See https://github.com/snapshot-labs/snapshot/blob/514f12c84d5986ef86f87193057d06d9b5c73353/src/views/ProfileAbout.vue#L59

The DNS check seems to be very hard to check live

ChaituVR (approved):
utAck

ChaituVR:
Don't merge until we merge stamp

PR #860, opened 2024-06-04
fix: return no spaces when controller is empty

wa0x6e:
Return an empty `spaces` result when the `controller` filter is empty

```graphql
query Spaces {
  spaces(
    first: 20,
    skip: 0,
    orderBy: "created",
    orderDirection: desc,
    where: { controller: "" }
  ) {
    id
  }
}
```

Should return 

```json
{
  "data": {
    "spaces": []
  }
}
```

and not all spaces.

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -180,7 +180,7 @@ export async function fetchSpaces(args) {
 
   const fields = { id: 'string', created: 'number' };
 
-  if (Object.hasOwnProperty.bind(where)('controller')) {
+  if ('controller' in where) {
     if (!where.controller) return [];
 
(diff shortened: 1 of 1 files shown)


src/graphql/helpers.ts
```diff
@@ -180,7 +180,9 @@ export async function fetchSpaces(args) {
 
   const fields = { id: 'string', created: 'number' };
 
-  if (where.controller) {
+  if (Object.hasOwnProperty.bind(where)('controller')) {
```
ChaituVR, line 183:
```suggestion
  if ('controller' in where) {
```

ChaituVR (changes requested):

ChaituVR (approved):
tAck

PR #861, opened 2024-06-05
fix: fix invalid returned keyname

wa0x6e:
Fix invalid returned keyname, should be `items` instead of `spaces`


ChaituVR (approved):
utAck

PR #869, opened 2024-06-17
fix: Increase statements id length

ChaituVR:
Same as https://github.com/snapshot-labs/snapshot-sequencer/pull/384


wa0x6e (approved):
utAc

PR #876, opened 2024-06-24
feat: return lens and farcaster from user

wa0x6e:
Related to https://github.com/snapshot-labs/sx-monorepo/pull/419

Return the new `lens` and `farcaster` properties from User

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -528,6 +528,8 @@ type User {
   cover: String
   github: String
   twitter: String
+  lens: String;
+  farcaster: String;
   created: Int!
 }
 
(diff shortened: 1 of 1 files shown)


src/graphql/schema.gql
```diff
@@ -528,6 +528,8 @@ type User {
   cover: String
   github: String
   twitter: String
+  lens: String;
+  farcaster: String;
```
ChaituVR, line 532:
```suggestion
  lens: String
  farcaster: String
```

ChaituVR (changes requested):

ChaituVR (approved):
utAck

PR #878, opened 2024-07-01
fix: handle lowercased addresses in inputs

ChaituVR:
- After changing collate of tables we have to support lowercased address like before, else many integrator's existing code may break


wa0x6e:
What about:

- messages:address
- alias:address/alias
- follows:follower
- user

ChaituVR:
- messages: Do we also want to change the collate of the `messages` table? i think it is not disturbing anything but would nice to keep everything same (but it may take a lot of time, we may have to do same as votes table)
- alias, follows, user: updated code ✅

wa0x6e (approved):
utAck

PR #880, opened 2024-07-04
fix: add support for starknet address

wa0x6e:
Fixes https://github.com/snapshot-labs/pitches/issues/83

This PR introduces some changes in order to accept starknet address as input (in `where` and `id` fields)

### Changes

- feat: add `starknetAddress` to `where` fields accepted values
- feat: allow passing an array to `where` values, so we can format using `evmAddress`, or `['evmAddress', 'starknetAddress']`
- refactor: rename `EVMAddress` to `evmAddress` in `where` fields formatter
- refactor: remove `@ethersproject/address` as dependency, will use the formatter exported by snapshot.js

### How to tests

- `yarn link` this branch of the SDK https://github.com/snapshot-labs/snapshot.js/pull/1032
- `yarn dev`
- you should be able to query fields by starknet address (e.g using a non padded starknet address or checksummed address, address will always be converted to padded/lowercased address before sent to mysql)

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -1,6 +1,13 @@
-import utils from '@snapshot-labs/snapshot.js/src/utils';
+import {
+  getFormattedAddress,
+  isEvmAddress,
+  isStarknetAddress
+} from '@snapshot-labs/snapshot.js/src/utils';
 import graphqlFields from 'graphql-fields';
 import fetch from 'node-fetch';
+import castArray from 'lodash/castArray';
+import intersection from 'lodash/intersection';
+import uniq from 'lodash/uniq';
 import { jsonParse } from '../helpers/utils';
 import { spacesMetadata } from '../helpers/spaces';
 import db from '../helpers/mysql';
@@ -115,6 +122,29 @@ export function formatSpace({
   return space;
 }
 
+export function formatAddresses(
+  addresses: string[],
+  types: ('evmAddress' | 'starknetAddress')[]
+) {
+  return addresses
+    .map(address => {
+      const results: string[] = [];
+
+      if (types.includes('evmAddress') && isEvmAddress(address)) {
+        results.push(getFormattedAddress(address, 'evm'));
+      } else if (
+        types.includes('starknetAddress') &&
+        isStarknetAddress(address)
+      ) {
+        results.push(getFormattedAddress(address, 'starknet'));
+      }
+
+      return results;
+    })
+    .flat()
+    .filter(Boolean);
+}
+
 export function buildWhereQuery(
   fields: Record<string, string | string[]>,
   alias: string,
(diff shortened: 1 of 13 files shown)


src/graphql/helpers.ts
```diff
@@ -31,7 +31,15 @@ export function formatAddress(address: string): string {
   try {
     return getAddress(address);
   } catch (error) {
-    // return same address for now, but return error in the future, if address is not EVM or sn address
+    // Pad starknet address to 64 characters
```
bonustrack, line 34: Can we support Starknet address without going through `getAddress` function before?

wa0x6e: Yes, I plan to move this function to the SDK, so we can reuse it inside sequencer and hub, with a more robust address detection

src/graphql/helpers.ts
```diff
@@ -1,6 +1,13 @@
-import { getAddress } from '@ethersproject/address';
+import {
+  getFormattedAddress,
+  isEvmAddress,
+  isStarknetAddress
+} from '@snapshot-labs/snapshot.js/src/utils';
```
ChaituVR, line 5: I think we should import like everywhere
Untitled 5

src/graphql/operations/user.ts
```diff
@@ -1,13 +1,18 @@
...
+import { formatUser, formatAddresses, PublicError } from '../helpers';
 
 export default async function (parent, args) {
-  const id = formatAddress(args.id);
+  const addresses = formatAddresses(
+    [args.id],
+    ['evmAddress', 'starknetAddress']
+  );
+  if (!addresses.length) throw new PublicError('Invalid address');
+
```
ChaituVR, line 12: Feel too confusing, it can't autoformat address.?

ChaituVR: i think we should let `formatAddresses` take care of types and errors

ChaituVR (changes requested):

src/graphql/helpers.ts
wa0x6e: Weird, I don't have this error on the console. What version of node are you using ?

src/graphql/operations/user.ts
wa0x6e: I we throw the error from `formatAddress`, we can not have a detailed error message explaining which field the error is coming from

src/graphql/helpers.ts
```diff
@@ -124,27 +118,67 @@ export function formatSpace({
...
+        types.includes('starknetAddress') &&
+        snapshot.utils.isStarknetAddress(address)
+      ) {
+        results.push(snapshot.utils.getFormattedAddress(address, 'starknet'));
+      }
+
+      return results;
+    })
+    .flat()
+    .filter(Boolean);
```
ChaituVR, line 147: I think it will be same with out these 🤔 
```suggestion
    });
```

src/graphql/helpers.ts
```diff
@@ -124,27 +118,67 @@ export function formatSpace({
...
-      }
+
+      conditions.forEach(condition => {
+        const key = `${field}${condition}`;
+
+        if (!where[key]) return;
+
+        const formattedAddresses = uniq(
+          formatAddresses(castArray(where[key]), arrayType)
+        );
```
ChaituVR, line 171: I think even if there is at least one invalid addresses, we should return error

ChaituVR (changes requested):

src/graphql/helpers.ts
ChaituVR: maybe we can check length of where[key] and

ChaituVR (approved):
tAck

added few comments but not too important

PR #881, opened 2024-07-10
fix: honor network when looking up domains controller

wa0x6e:
Fixes # .

- 
- 
-

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -431,7 +431,8 @@ async function getControllerDomains(address: string): Promise<string[]> {
       },
       body: JSON.stringify({
         method: 'lookup_domains',
-        params: address
+        params: address,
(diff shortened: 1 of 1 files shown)


src/graphql/helpers.ts
```diff
@@ -431,7 +431,8 @@ async function getControllerDomains(address: string): Promise<string[]> {
       },
       body: JSON.stringify({
         method: 'lookup_domains',
-        params: address
+        params: address,
+        network: process.env.NETWORK === 'mainnet' ? '1' : '11155111'
```
ChaituVR, line 435: It is actually `livenet` https://hub.snapshot.org/api/, better to look for `testnet`
```suggestion
        network: network === 'testnet' ? '11155111' : '1'
```

ChaituVR (changes requested):

PR #885, opened 2024-07-12
fix: fix duplicates items on leaderboard pagination

wa0x6e:
Will fix the offchain part for https://github.com/snapshot-labs/sx-monorepo/issues/480

Paginating through the leaderboard is returning duplicates items.

## 💊 Fixes / Solution

Adding more ordering fields

## 🚧 Changes

- Always sort by 3 fields (`proposal_count`, `vote_count` and `user`), to avoid duplicates when multiple rows have the same counter
- The `orderBy` field will only move the given order field to the beginning

## 🛠️ Tests

```graphql
query {
  leaderboards(
    first: 20
    skip: 0
    orderBy: "proposal_count"
    orderDirection: desc
    where: {
    space: "safe.eth"
  }
  ) {
    user
    space
    proposalsCount
    votesCount
  }
}
```

Increment the `skip` by 20 will not returning duplicates items from previous page

### DIFF
--- a/src/graphql/operations/leaderboards.ts
+++ b/src/graphql/operations/leaderboards.ts
@@ -1,3 +1,4 @@
+import uniq from 'lodash/uniq';
 import db from '../../helpers/mysql';
 import log from '../../helpers/log';
 import { buildWhereQuery, checkLimits } from '../helpers';
@@ -8,6 +9,9 @@ export default async function (parent, args) {
 
   checkLimits(args, 'leaderboards');
 
+  const ORDER_FIELDS = ['vote_count', 'proposal_count', 'user'];
+  const DEFAULT_ORDER_FIELD = 'vote_count';
+
   const fields = {
     user: 'EVMAddress',
     space: 'string',
@@ -15,13 +19,16 @@ export default async function (parent, args) {
     proposal_count: 'number'
   };
   const whereQuery = buildWhereQuery(fields, 'l', where);
-  const defaultOrder = 'votesCount DESC, proposalsCount DESC';
 
-  const orderBy = Object.keys(fields).includes(args.orderBy)
+  const orderBy = ORDER_FIELDS.includes(args.orderBy)
     ? args.orderBy
-    : null;
+    : DEFAULT_ORDER_FIELD;
   let orderDirection = (args.orderDirection || 'desc').toUpperCase();
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
+
+  const orderQuery = uniq([orderBy, ...ORDER_FIELDS])
+    .map((field: string) => `l.${field} ${orderDirection}`)
+    .join(', ');
 
   const query = `
     SELECT l.*,
@@ -30,9 +37,7 @@ export default async function (parent, args) {
       l.last_vote as lastVote
     FROM leaderboard l
     WHERE 1=1 ${whereQuery.query}
-    ORDER BY ${
-      orderBy ? `l.${orderBy} ${orderDirection}` : defaultOrder
-    } LIMIT ?, ?
+    ORDER BY ${orderQuery} LIMIT ?, ?
   `;
 
   try {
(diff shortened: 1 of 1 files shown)


src/graphql/operations/leaderboards.ts
```diff
@@ -8,31 +9,35 @@ export default async function (parent, args) {
...
 
+  const orderQuery = uniq([orderBy, ...ORDER_FIELDS])
+    .map((field: string) => `l.${field} ${orderDirection}`)
+    .join(', ');
+
   const query = `
     SELECT l.*,
       l.vote_count as votesCount,
       l.proposal_count as proposalsCount,
       l.last_vote as lastVote
```
ChaituVR, line 37: Sorry for not noticing it earlier, but I think it is not very clear that we use underscore in inputs and camelCase on outputs, wyt? 🤔 

because `orderBy` is usually `string` and we may pick it from the output schema

src/graphql/operations/leaderboards.ts
```diff
@@ -8,31 +9,35 @@ export default async function (parent, args) {
...
 
-  const orderBy = Object.keys(fields).includes(args.orderBy)
+  const orderBy = ORDER_FIELDS.includes(args.orderBy)
     ? args.orderBy
-    : null;
+    : DEFAULT_ORDER_FIELD;
   let orderDirection = (args.orderDirection || 'desc').toUpperCase();
   if (!['ASC', 'DESC'].includes(orderDirection)) orderDirection = 'DESC';
 
+  const orderQuery = uniq([orderBy, ...ORDER_FIELDS])
```
ChaituVR, line 29: why to include other `ORDER_FIELDS` 🤔

src/graphql/operations/leaderboards.ts
ChaituVR: also input is `vote` count while output is `votes` count

src/graphql/operations/leaderboards.ts
ChaituVR: weird that it is taking more time without it :D

ChaituVR (approved):
tAck

left some comments but they are not blocking this

src/graphql/operations/leaderboards.ts
wa0x6e: I use `proposalsCount` as output, to follow the convention in Space. It is camel case, and plural

src/graphql/operations/leaderboards.ts
```diff
@@ -8,31 +9,35 @@ export default async function (parent, args) {
 
   checkLimits(args, 'leaderboards');
 
+  const ORDER_FIELDS = ['vote_count', 'proposal_count', 'user'];
```
ChaituVR, line 12: I don't think we would ever use `user` in orderby, I think we may use the last vote to sort recently active users

wa0x6e: We could add `last_vote` and use it internally, but `user` has been added to keep follow sx api

https://api-1.snapshotx.xyz/?explorerURLState=N4IgJg9gxgrgtgUwHYBcQC4QEcYIE4CeABMADpJFEDOADgIZQIAUAlmOkaSAigBboAPIQK4BKEuUqU2kogF9yckHKA

src/graphql/operations/leaderboards.ts
wa0x6e: Including all fields to avoid duplication, like when we have multiple voters with 1 vote and 0 proposal, we should still have a fallback field to sort with

src/graphql/operations/leaderboards.ts
wa0x6e: We have the same issue with `User`. We use `proposal_count` in where and order, and it return `votesCount`.

Not sure why we’re using a mix of underscore and camel case for output, but I just followed what we’re already using for counter fields

src/graphql/operations/leaderboards.ts
ChaituVR: i see, got it. 

yea i think better to add last vote and space, but not important for now. feel free to merge

wa0x6e: - `last_vote` added to the `ORDER BY`
- Add back all other fields (`space`) to the order by whitelist

ChaituVR (approved):
utAck

PR #886, opened 2024-07-14
fix: user query should return null if user does not exist

wa0x6e:

Current code only return result when the address exist on user (and optionally on leaderboard with `LEFT JOIN`), so querying an user with an address not existing on the users table, but existing in the leaderboard table returns the following error

```graphql
query User {
  user(id: "0x00Cef6B61fd57d481e569C64Ac5280736023de7F") {
    id
  }
}
```

```graphql
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field User.id.",
      "locations": [
        {
          "line": 4,
          "column": 5
        }
      ],
      "path": [
        "user",
        "id"
      ]
    }
  ],
  "data": {
    "user": null
  }
}
```

This is due to the query always returning something, but without `users.id`, because of `COALESCE()` on the leaderboard counter columns.

## 💊 Fixes / Solution

Return `null` when user does not exist, instead of an user without ID

## 🚧 Changes

- When `JOIN` with leaderboard, return result only when the user exist on the `users` table

## 🛠️ Tests

Test with the following query, on a repo connected to the prod database (for the data)

```graphql
query User {
  user(id: "--ADDRESS--") {
    id
    name
    about
    avatar
    cover
    github
    twitter
    lens
    farcaster 
    proposalsCount
    votesCount
    lastVote
  }
}
```

with the following addresses:

- only on leaderboard: `0x0000000007c96AD6c9835204303c8fB35E48Ffc3`
- only on user: `0x2A7e46945277059D57a795CbecDb92DC29db69A6`
- on both tables: `0x04DB1bB49b7fFBcEC574f34D29c3153953890352`

It should gives similar results on a `users` query:

```graphql

query Users {
  users(where: { id_in: ["0x0000000007c96AD6c9835204303c8fB35E48Ffc3", "0x2A7e46945277059D57a795CbecDb92DC29db69A6", "0x04DB1bB49b7fFBcEC574f34D29c3153953890352", "0x04DB1bB49b7fFBcEC574f34D29c3153953890000"] }) {
    id
    name
    about
    avatar
    cover
    github
    twitter
    lens
    farcaster 
    proposalsCount
    votesCount
    lastVote
  }
}
```

The query should not return any results for `0x0000000007c96AD6c9835204303c8fB35E48Ffc3` (only leaderboard)

### DIFF
--- a/src/graphql/operations/user.ts
+++ b/src/graphql/operations/user.ts
@@ -6,19 +6,38 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
 export default async function (parent, args) {
   const id = formatAddress(args.id);
   const query = `
-    SELECT
-      u.*,
-      SUM(l.vote_count) as votesCount,
-      SUM(l.proposal_count) as proposalsCount,
-      MAX(l.last_vote) as lastVote
-    FROM users u
-    LEFT JOIN leaderboard l ON l.user = u.id
-    WHERE id = ?
+    SELECT * FROM (
+      SELECT
+        u.*,
+        u.id as userId,
+        COALESCE(SUM(l.vote_count), 0) as votesCount,
+        COALESCE(SUM(l.proposal_count), 0) as proposalsCount,
+        MAX(l.last_vote) as lastVote
+      FROM users u
+      LEFT JOIN leaderboard l ON l.user = u.id
+      GROUP BY u.id
+      UNION
+      SELECT
+        u.*,
+        l.user as userId,
+        COALESCE(SUM(l.vote_count), 0) as votesCount,
+        COALESCE(SUM(l.proposal_count), 0) as proposalsCount,
+        MAX(l.last_vote) as lastVote
+      FROM users u
+      RIGHT JOIN leaderboard l ON l.user = u.id
+      GROUP BY l.user
+    ) AS t
+    WHERE t.userId = ?
     LIMIT 1`;
   try {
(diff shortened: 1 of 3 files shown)


src/graphql/operations/user.ts
```diff
@@ -6,19 +6,35 @@ import { capture } from '@snapshot-labs/snapshot-sentry';
...
+        l.user as userId,
+        COALESCE(SUM(l.vote_count), 0) as votesCount,
+        COALESCE(SUM(l.proposal_count), 0) as proposalsCount,
+        MAX(l.last_vote) as lastVote
+      FROM users u
+      RIGHT JOIN leaderboard l ON l.user = u.id
+      GROUP BY l.user
+    ) AS t
+    WHERE t.userId = ?
     LIMIT 1`;
```
ChaituVR, line 31: Getting too big, can be confusing, is it possible to get both users and the leaderboard and merge them?

ChaituVR (changes requested):

src/graphql/operations/user.ts
wa0x6e: This is what the query is doing, merge both tables then query it.

This is the mysql way of doing a `FULL OUTER JOIN`, since it does not support it.

wa0x6e:
`proposalsCount` and `votesCount` return `null` when user don't exist on the leaderboard

wa0x6e:
See #891

PR #887, opened 2024-07-14
fix: optimize spaces helpers 

wa0x6e:
Use the new `vote_count`, `proposal_count` and `follower_count` columns from the spaces table to retrieve total number of items per space, instead of using a `COUNT(id) as COUNT` select.

Also some refactoring to add types, and cleaning.

The whole script now run in 40 seconds (instead of > 10 minutes), with 95% of the time spend on loading all spaces.

### How to test

1. Connect your snapshot UI to a hub with running branch
2. Front page should load the same spaces, in the same order

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -133,7 +133,7 @@ async function loadSpaces() {
 
   while (true) {
     const query = `
-      SELECT id, settings, flagged, verified, turbo, hibernated
+      SELECT id, settings, flagged, verified, turbo, hibernated, follower_count, proposal_count, vote_count
       FROM spaces
       WHERE deleted = 0
       ORDER BY id ASC
@@ -151,7 +151,10 @@ async function loadSpaces() {
           flagged: space.flagged === 1,
           verified: space.verified === 1,
           turbo: space.turbo === 1,
-          hibernated: space.hibernated === 1
+          hibernated: space.hibernated === 1,
+          follower_count: space.follower_count,
+          vote_count: space.vote_count,
+          proposal_count: space.proposal_count
         }
       ])
     );
@@ -263,7 +266,7 @@ async function loadSpacesMetrics() {
 
 export async function getSpace(id: string) {
   const query = `
-    SELECT settings, flagged, verified, turbo, hibernated, deleted
+    SELECT settings, flagged, verified, turbo, hibernated, deleted, follower_count, proposal_count, vote_count
     FROM spaces
     WHERE id = ?
     LIMIT 1`;
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -6,6 +6,18 @@ import log from './log';
 import { capture } from '@snapshot-labs/snapshot-sentry';
 
 const RUN_INTERVAL = 120e3;
+const TEST_STRATEGIES = [
+  'ticket',
+  'api',
+  'api-v2',
+  'api-post',
+  'api-v2-override'
+];
+const TESTNET_NETWORKS = (
+  Object.values(networks) as { testnet?: boolean; key: string }[]
+)
+  .filter(network => network.testnet)
+  .map(network => network.key);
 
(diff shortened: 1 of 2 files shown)


src/helpers/spaces.ts
```diff
@@ -134,28 +134,40 @@ async function loadSpaces() {
...
 
 async function getFollowers() {
   const query = `
-    SELECT space, COUNT(id) as count,
-    count(IF(created > (UNIX_TIMESTAMP() - 604800), 1, NULL)) as count_7d
-    FROM follows GROUP BY space
+    SELECT
+      space,
+      spaces.follower_count AS count,
+      COUNT(IF(created > (UNIX_TIMESTAMP() - 604800), 1, NULL)) AS count_7d
```
ChaituVR, line 167: I get this error `Column 'created' in field list is ambiguous` , I think it should be `follows.created`, will add a fix

src/helpers/spaces.ts
```diff
@@ -134,28 +134,40 @@ async function loadSpaces() {
...
-    SELECT space, COUNT(id) as count,
-    count(IF(created > (UNIX_TIMESTAMP() - 604800), 1, NULL)) as count_7d
-    FROM votes GROUP BY space
+    SELECT
+      space,
+      spaces.vote_count as count,
+      COUNT(IF(v.created > (UNIX_TIMESTAMP() - 604800), 1, NULL)) AS count_7d
+    FROM votes v
+    JOIN spaces ON spaces.id = space
+    GROUP BY space
```
ChaituVR, line 157: This one is taking more than 20 mins (still running) but the previous one took around 12.5 mins

ChaituVR (changes requested):

ChaituVR:
I think solution should be to ignore total counts here in these queries, but use space counts directly on `mapSpaces` in `src/helpers/spaces.ts`

src/helpers/spaces.ts
ChaituVR: It took 24 mins

wa0x6e: How can a simple read on a column `vote_count` be slower than a `COUNT(id)` ?

wa0x6e: Maybe the slowness is not related to the update, and the whole query itself does not take same time on each run ? (sometime 12min, sometime 20min, sometime 30min) ?

src/helpers/spaces.ts
```diff
@@ -134,28 +134,40 @@ async function loadSpaces() {
...
-    SELECT space, COUNT(id) as count,
-    count(IF(created > (UNIX_TIMESTAMP() - 604800), 1, NULL)) as count_7d
-    FROM follows GROUP BY space
+    SELECT
+      space,
+      spaces.follower_count AS count,
+      COUNT(IF(f.created > (UNIX_TIMESTAMP() - 604800), 1, NULL)) AS count_7d
+    FROM follows f
+    JOIN spaces ON spaces.id = space
+    GROUP BY space
```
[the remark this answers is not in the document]
wa0x6e: Let me try an `explain`, it's very weird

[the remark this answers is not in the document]
wa0x6e: Ah, `JOIN` was the expensive part. I may have found a better query, which is almost instantaneous 😎😎

src/helpers/spaces.ts
ChaituVR: I think slowness could be because of JOIN spaces ON spaces.id = space, adding spaces to all votes

ChaituVR: we already have all space data in `mapSpaces` so we don't need to get this data here, we can just get `7d` counts here and we should be good

wa0x6e: Yes, i found a better query, everything will be done in less than 10s

ChaituVR (approved):
tAck 👍

PR #891, opened 2024-07-16
fix: separate leaderboard query on users query

ChaituVR:
Continued from https://github.com/snapshot-labs/snapshot-hub/pull/886

### Summary:

- Separate query to leaderboard table 
- user query will call users query instead of duplicate code
- Remove unused  filters on users query
- `created` can be null

### How to test:
Same as https://github.com/snapshot-labs/snapshot-hub/pull/886#issue-2407239841

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -363,7 +363,6 @@ export async function handleRelatedSpaces(info: any, spaces: any[]) {
 export function formatUser(user) {
   const profile = jsonParse(user.profile, {});
   delete user.profile;
-
   return {
     ...user,
     ...profile
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -124,11 +124,20 @@ export function formatSpace({
   return space;
 }
 
-export function buildWhereQuery(fields, alias, where) {
+export function buildWhereQuery(
+  fields,
+  alias,
+  where,
+  options: { aliases?: Record<string, string> } = {}
+) {
   let query: any = '';
   const params: any[] = [];
 
   Object.entries(fields).forEach(([field, type]) => {
+    const aliasedFieldName = `${alias}.${
+      (options.aliases || {})[field] || field
+    }`;
+
     if (type === 'EVMAddress') {
       const conditions = ['', '_not', '_in', '_not_in'];
       try {
@@ -145,20 +154,20 @@ export function buildWhereQuery(fields, alias, where) {
       }
     }
     if (where[field] !== undefined) {
-      query += `AND ${alias}.${field} = ? `;
+      query += `AND ${aliasedFieldName} = ? `;
       params.push(where[field]);
     }
 
     const fieldNot = where[`${field}_not`];
     if (fieldNot !== undefined) {
-      query += `AND ${alias}.${field} != ? `;
+      query += `AND ${aliasedFieldName} != ? `;
       params.push(fieldNot);
     }
 
     const fieldIn = where[`${field}_in`];
     if (Array.isArray(fieldIn)) {
       if (fieldIn.length > 0) {
-        query += `AND ${alias}.${field} IN (?) `;
+        query += `AND ${aliasedFieldName} IN (?) `;
         params.push(fieldIn);
       } else {
         query += 'AND 1=0 ';
@@ -168,7 +177,7 @@ export function buildWhereQuery(fields, alias, where) {
     const fieldNotIn = where[`${field}_not_in`];
     if (Array.isArray(fieldNotIn)) {
       if (fieldNotIn.length > 0) {
-        query += `AND ${alias}.${field} NOT IN (?) `;
+        query += `AND ${aliasedFieldName} NOT IN (?) `;
         params.push(fieldNotIn);
       } else {
         query += 'AND 1=0 ';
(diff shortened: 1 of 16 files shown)


src/graphql/helpers.ts
```diff
@@ -363,6 +363,7 @@ export async function handleRelatedSpaces(info: any, spaces: any[]) {
 export function formatUser(user) {
   const profile = jsonParse(user.profile, {});
   delete user.profile;
+
```
ChaituVR, line 366:
```suggestion
```

wa0x6e:
Query for single user is not working, always returning the error 

```
{
  "errors": [
    {
      "message": "Cannot return null for non-nullable field User.id.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ],
      "path": [
        "user",
        "id"
      ]
    }
  ],
  "data": {
    "user": null
  }
}
```

wa0x6e:
What I meant by possible weird returned results depending on filter is:

```
query User {
    users(where: {id_in: ["0x0000000007c96AD6c9835204303c8fB35E48Ffc3", "0x2A7e46945277059D57a795CbecDb92DC29db69A6", "0x04DB1bB49b7fFBcEC574f34D29c3153953890352", "0x00Cef6B61fd57d481e569C64Ac5280736023de7F"] }) {
    id
    proposalsCount
    votesCount
  }
}
```

And 

```
query User {
    users(where: {id_in: ["0x0000000007c96AD6c9835204303c8fB35E48Ffc3", "0x2A7e46945277059D57a795CbecDb92DC29db69A6", "0x04DB1bB49b7fFBcEC574f34D29c3153953890352", "0x00Cef6B61fd57d481e569C64Ac5280736023de7F"], created_lte: 0 }) {
    id
    proposalsCount
    votesCount
  }
}
``` 

do not return the same results

ChaituVR:
It is a different issue,  you are using `created_lte: 0`
which never work somehow, for example try:
```GraphQL
query Proposals{
  proposals(first: 10, where:{created_lte:0}) {
    id
    created
  }
}
```

If you pass a valid value, then it returns correct values, it filters out Users with created `null`

```Graphql
query User {
    users(where: {id_in: ["0x0000000007c96AD6c9835204303c8fB35E48Ffc3", "0x2A7e46945277059D57a795CbecDb92DC29db69A6", "0x04DB1bB49b7fFBcEC574f34D29c3153953890352", "0x00Cef6B61fd57d481e569C64Ac5280736023de7F"], created_lte: 1651651402 }) {
    id
    proposalsCount
    votesCount
    created
  }
}
```

ChaituVR:
> Query for single user is not working, always returning the error

Fixed `user` query, it was due to last minute refactoring 😅

src/graphql/operations/users.ts
```diff
@@ -16,6 +16,8 @@ export default async function (parent, args) {
   const whereQuery = buildWhereQuery(fields, 'u', where);
   const queryStr = whereQuery.query;
   const params: any[] = whereQuery.params;
+  const ids = (where.id_in || [where.id]).filter((id: any) => id);
```
wa0x6e, line 19:
```suggestion
  const ids = (where.id_in || [where.id]).filter(Boolean);
```

src/graphql/operations/users.ts
```diff
@@ -16,6 +16,8 @@ export default async function (parent, args) {
...
+  if (Object.keys(where).length > 1) ids.length = 0;
```
wa0x6e, line 20: Don't get this part

src/graphql/operations/users.ts
```diff
@@ -26,20 +28,41 @@ export default async function (parent, args) {
 
   const query = `
     SELECT
-      u.*,
-      SUM(l.vote_count) as votesCount,
-      SUM(l.proposal_count) as proposalsCount,
-      MAX(l.last_vote) as lastVote
+      u.*
```
wa0x6e, line 31: Maybe keep the leaderboard `JOIN` on this query ?

So that if someone requests an user present in both users and leaderboard tables, we can do everything with this single query.

Reserve the second query only in case address exists only on leaderboard table, and not users.

wa0x6e (changes requested):

src/graphql/operations/users.ts
ChaituVR: If there are more than 1 filters, users in `leaderboard` table will be ignored, to give more importance to these filters, 

If either `id_in` or `id` exist in filters, we fetch users from both tables

wa0x6e: `ids` is an array. You're manipulating the `length` property of an array, and assigning it a new value ?

ChaituVR: Yes yes, this will make `ids = []` an empty array if there are other filters in `where`

src/graphql/operations/users.ts
ChaituVR: Changed it

src/graphql/operations/users.ts
wa0x6e: Did not know that way of clearing an array ... the wonder of JS

src/graphql/operations/users.ts
wa0x6e: The `INNER JOIN` will only consider addresses that exists in both users and leaderboard table.

Better to user `LEFT JOIN`, so we catch addresses that exists on both tables and only on users table on the first query (with some `COALESCE` on the counters), so the second query will take care only of addresses that do not exist at all on the users table.

src/graphql/operations/users.ts
```diff
@@ -27,19 +29,52 @@ export default async function (parent, args) {
...
+      `,
+        [usersWithOutCreated]
+      );
+      counts.forEach((count: any) => {
+        const user = users.find((u: any) => u.id === count.user);
+        if (user) {
+          user.votesCount = count.votesCount;
+          user.proposalsCount = count.proposalsCount;
+          user.lastVote = count.lastVote;
+        }
```
wa0x6e, line 74:
```suggestion
        user.votesCount = count.votesCount;
        user.proposalsCount = count.proposalsCount;
        user.lastVote = count.lastVote;
        
```

wa0x6e (changes requested):

src/graphql/operations/users.ts
wa0x6e: The case where `user` does not exist will never happen, as we get the ID from this object

wa0x6e (approved):
tAck

PR #893, opened 2024-07-17
fix: Use lowercased proposal title to search

### LINKED ISSUE #892: bug: search for proposal is case sensitive
Currently it's really hard to search for a proposal on snapshot because it's case sensitive, which was not the case previously, this is most likely due to recent change with collation. We should fix this

ChaituVR:
Fixes #892 

## Summary:
- Use lowercased titles to search with `LOWER`

How to test:
- Query using
```Graphql
query Proposals {
  proposals(
    where: {title_contains: "ARFC-ADDENDUM"}
  ) {
    id
    title
  }
}
```
- Should not work before but should work after

### DIFF
--- a/src/graphql/operations/proposals.ts
+++ b/src/graphql/operations/proposals.ts
@@ -41,8 +41,8 @@ export default async function (parent, args) {
 
   let searchSql = '';
   if (where.title_contains) {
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

PR #908, opened 2024-08-13
fix: ranking query should also return spaces with no popularity

ChaituVR:
### Summary
Some spaces don't have popularity when they are created, these should not be ignored by search
image
image

### How to test 
- Try query before and after fix, it should return space 
```graphql
{
  ranking(where: {search: "ITP DAO"}) {
    items {
      id
    }
  }
}

```


wa0x6e (approved):
utAck

PR #911, opened 2024-08-21
DAOstar Collab: Improve ProposalsURI - Publish more info for proposals

Rashmi-278:
Fixes # .

- Add Content URI Field, which includes Author, IPFS hash, Body, Link of the proposal
- Add Discussion URI Field, which includes the forum discussion link

Reference Issues:
- https://github.com/metagov/daostar/issues/203
- https://github.com/metagov/daostar/issues/239

### DIFF
--- a/src/eip4824.ts
+++ b/src/eip4824.ts
@@ -70,7 +70,7 @@ router.get('/:space/proposals', async (req, res) => {
     if (!space.verified) return res.status(400).json({ error: 'INVALID' });
 
     proposals = await db.queryAsync(
-      'SELECT id, title, start, end FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
+      'SELECT id, title, start, end, body, author, ipfs, link, discussion FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
       [id]
     );
   } catch (e) {
@@ -101,6 +101,7 @@ router.get('/:space/proposals', async (req, res) => {
     proposals
   });
 });
+
 
 router.get('/:space/activities', async (req, res) => {
   const id = req.params.space;
--- a/src/eip4824.ts
+++ b/src/eip4824.ts
@@ -70,7 +70,7 @@ router.get('/:space/proposals', async (req, res) => {
     if (!space.verified) return res.status(400).json({ error: 'INVALID' });
 
     proposals = await db.queryAsync(
-      'SELECT id, title, start, end, body, author, ipfs, link, discussion FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
+      'SELECT * FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
       [id]
     );
   } catch (e) {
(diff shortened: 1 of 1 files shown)


src/eip4824.ts
```diff
@@ -82,6 +82,13 @@ router.get('/:space/proposals', async (req, res) => {
     type: 'proposal',
     id: proposal.id,
     name: proposal.title,
+    contentURI: {
+      body: proposal.body,
+      author: proposal.author,
+      ipfs: proposal.ipfs,
+      link: proposal.link
+    },
+    discussionURI: proposal.discussion,
```
ChaituVR, line 91: I get `"contentURI": {},` and no `discussionURI` field in response

ChaituVR (changes requested):

src/eip4824.ts
Rashmi-278: It should be fixed now, can you check?

src/eip4824.ts
```diff
@@ -95,6 +102,7 @@ router.get('/:space/proposals', async (req, res) => {
   });
 });
 
+
```
ChaituVR, line 105:
```suggestion
```

src/eip4824.ts
```diff
@@ -82,6 +82,13 @@ router.get('/:space/proposals', async (req, res) => {
     type: 'proposal',
     id: proposal.id,
     name: proposal.title,
+    contentURI: {
+      body: proposal.body,
+      author: proposal.author,
+      ipfs: proposal.ipfs,
+      link: proposal.link
```
ChaituVR, line 89: There is no link field on proposal 🤔 what should be the value

bonustrack: The snapshot URL of the proposal, we have this on the hub API

Rashmi-278: Link of the Snapshot Proposal

`"link": "https://snapshot.org/#/opcollective.eth/proposal/0xc9c0a739a29ce6477708f49a2985e4c8cb2e4ba0be1d3761541ed189a9c341e1",`

I found this from GraphQL API,

ChaituVR: @Rashmi-278 no access for me to change it,

Could you change these:
- Remove `link` from the above query
- You can use the same logic as https://github.com/metagov/snapshot-hub/blob/pr/Rashmi-278/911/src/graphql/helpers.ts#L401-L402

Rashmi-278: I face 404 on the above link, do you mean to remove link from query and `Query *` ?

src/eip4824.ts
```diff
@@ -70,7 +70,7 @@ router.get('/:space/proposals', async (req, res) => {
     if (!space.verified) return res.status(400).json({ error: 'INVALID' });
 
     proposals = await db.queryAsync(
-      'SELECT id, title, start, end FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
+      'SELECT id, title, start, end, body, author, ipfs, link, discussion FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
```
ChaituVR, line 73:
```suggestion
      'SELECT id, title, start, end, body, author, ipfs, discussion FROM proposals WHERE space = ? ORDER BY created DESC LIMIT 20',
```

PR #913, opened 2024-08-22
feat: add support for new source column in statements table

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/167

Add support for the new `source` column in the statements table

- Add ability to query statements by source (string and array)
- Can return `source` in Statement
- Update schema.sql to reflect new source column

### Note 

To be merged after https://github.com/snapshot-labs/snapshot-sequencer/pull/422


bonustrack (approved):
utACK

PR #914, opened 2024-08-23
feat: add `verified` field to SpacesWhere

### LINKED ISSUE #907: feat: add filter for verified spaces
We should have a filter to query only verified spaces

wa0x6e:
Closes #907

This PR add a `verified` filter to the spaces `where` argument

### Test 

```
query Spaces {
  spaces(where: { verified: true }) {
    id 
    name
  }
}
```

This should return a list of only verified spaces

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -234,6 +234,11 @@ export function buildWhereQuery(
         params.push(fieldLte);
       }
     }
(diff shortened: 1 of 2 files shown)


bonustrack (approved):
utACK, this was requested by DAOStar, we should ping them once merged.

PR #916, opened 2024-08-28
feat: add network to delegation portal

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/177

This PR will return the new `delegationNetwork` property in the Delegation Portal setting.

When the property does not exist (for old data), it will always return `1` (ethereum).

### Test 

This space does not have `delegationNetwork` set, and the hub will return `1`

```graphql
query Spaces {
  spaces (where: { id: "test.wa0x6e.eth"}  ){
    id
    delegationPortal {
      delegationApi
      delegationType
      delegationNetwork
      delegationContract
    }
  }
}
```


ChaituVR (approved):
tAck

Issue #760, opened 2023-11-27
bug: votes query return `null` value


ChaituVR:
Try the following query
```GraphQL
query Votes {
    votes(first: 1000, where: { voter: "0x24F15402C6Bb870554489b2fd2049A85d75B982f" }) {
      proposal {
        id
        space {
          id
        }
        ipfs
      }
    }
  }
  ```

It will return a response that contains this
![image](https://github.com/snapshot-labs/snapshot-hub/assets/15967809/59112840-34d3-47bd-a32a-cea857802ea9)

**Comment by wa0x6e** [comment]
I am not getting the same graphql response. Was the DB cleaned up in the meantime ?

**Comment by ChaituVR** [comment]
No change. can try again?

It is bit down in response, search for `null`
![image](https://github.com/snapshot-labs/snapshot-hub/assets/15967809/8241cefd-c4db-4edc-bb8e-8ba588cb4b60)

**Comment by wa0x6e** [comment]
How is that possible when we use `DELETE FROM votes WHERE proposal = ?;` when deleting proposals ?

Is it just bug, which can be resolve by consolidating the votes database ?

**Comment by wa0x6e** [comment]
This is a an issue with orphan proposals, where the proposals are not linked to a valid space anymore.

Either the space has been renamed, deleted, or the something's wrong happened during the migration.

You can see the list of proposals without any valid space with

```sql
SELECT space from proposals where space NOT IN (select id from spaces) group by space;
```

There are currently 81 proposals in that situation

PR #921, opened 2024-09-07
fix: Exclude votes without proposals

ChaituVR:
Temporary fix for #760 

#### How to test:

- Test with:
```GraphQL
query Votes {
    votes(first: 1000, where: { voter: "0x24F15402C6Bb870554489b2fd2049A85d75B982f" }) {
     id  
    proposal {
        id
        space {
          id
        }
        ipfs
      }
    }
  }
  
 ```
 Should return values without `null` in proposal
- 
```GraphQL
query { 
  vote(id: "0xd954ffa9c8766c7f1195f867522ac759d8f752781d811941c91e289032e7df90") {
    
    id
    proposal {
      id
    }
  }
}
```
Should return `null`

### DIFF
--- a/src/graphql/operations/votes.ts
+++ b/src/graphql/operations/votes.ts
@@ -44,6 +44,7 @@ async function query(parent, args, context?, info?) {
 
   const query = `
     SELECT v.* FROM votes v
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
utAck

PR #927, opened 2024-09-23
fix: fix issue where sub spaces querying return 20 results max

wa0x6e:
Fix an issue where querying multiple spaces, with a combined total of subspaces > 20, is returning 20 subspaces max


bonustrack (approved):
utACK

PR #937, opened 2024-10-02
fix: return snapshot as number

wa0x6e:
Closes https://github.com/snapshot-labs/workflow/issues/232

This PR returns the `snapshot` property as number (since in the table, it's an INT column, and on the front end, we're expecting a number)


ChaituVR (approved):
utAck

PR #944, opened 2024-10-10
fix: Use domain field instead of one in JSON

ChaituVR:
Related to https://github.com/snapshot-labs/workflow/issues/241

### Summery
- Use `domain` field instead of `domain` in settings JSON

### How to test:
- Change `domain` field and `settings.domain`
- Hub queries should return `domain` field instead of domain in JSON


wa0x6e (approved):
tAck

PR #950, opened 2024-10-24
fix: Remove proposal check in votes query 

ChaituVR:
Fixes https://github.com/snapshot-labs/snapshot-hub/issues/760

- We don't have votes with `null` proposal
- Removes check for proposal in votes query 

### How to test
- Use this query
```graphql
query ($spaceId: String!, $where: VoteWhere!, $orderDirection: OrderDirection!) {
  votes(where: $where, orderDirection: $orderDirection) {
    created
    voter
  }
  space(id: $spaceId) {
    admins
    members
    moderators
  }
}
```
Variables
```JSON
{
  "spaceId": "arbitrumfoundation.eth",
  "where": {
    "space": "arbitrumfoundation.eth",
    "created_gt": 0
  },
  "orderDirection": "DE"
}
```
- It should be super slow before fix and relatively fast after fix

### DIFF
--- a/src/graphql/operations/votes.ts
+++ b/src/graphql/operations/votes.ts
@@ -44,9 +44,8 @@ async function query(parent, args, context?, info?) {
 
   const query = `
     SELECT v.* FROM votes v
-    INNER JOIN proposals p ON v.proposal = p.id
     WHERE 1 = 1 ${queryStr}
-    ORDER BY ${orderBy} ${orderDirection}, v.id ASC LIMIT ?, ?
+    ORDER BY ${orderBy} ${orderDirection} LIMIT ?, ?
   `;
   params.push(skip, first);
   try {
--- a/src/graphql/operations/votes.ts
+++ b/src/graphql/operations/votes.ts
@@ -45,7 +45,7 @@ async function query(parent, args, context?, info?) {
   const query = `
     SELECT v.* FROM votes v
     WHERE 1 = 1 ${queryStr}
-    ORDER BY ${orderBy} ${orderDirection} LIMIT ?, ?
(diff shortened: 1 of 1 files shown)


bonustrack:
The id sorting is here because we've noticed that it can create duplicates and remove votes in pagination is there are multiple votes with the same timestamp

ChaituVR:
> removing the `JOIN...` is not enough to make the query fast ?

Yep, for example try running 
```SQL
SELECT v.* FROM votes v
    WHERE 1 = 1 AND v.space = 'arbitrumfoundation.eth'
    ORDER BY v.created DESC, v.id ASC LIMIT 0, 1
```
vs
```SQL
SELECT v.* FROM votes v
    WHERE 1 = 1 AND v.space = 'arbitrumfoundation.eth'
    ORDER BY v.created DESC LIMIT 0, 1
```

ChaituVR:
But anyway i will do seperate PR for that, lets merge that first.

src/graphql/operations/votes.ts
```diff
@@ -44,9 +44,8 @@ async function query(parent, args, context?, info?) {
 
   const query = `
     SELECT v.* FROM votes v
-    INNER JOIN proposals p ON v.proposal = p.id
     WHERE 1 = 1 ${queryStr}
-    ORDER BY ${orderBy} ${orderDirection}, v.id ASC LIMIT ?, ?
+    ORDER BY ${orderBy} ${orderDirection} LIMIT ?, ?
```
ChaituVR, line 48:
```suggestion
    ORDER BY ${orderBy} ${orderDirection}, v.id ASC LIMIT ?, ?
```

wa0x6e (approved):
utAck

PR #951, opened 2024-10-27
feat: return email subscription data in User

wa0x6e:
Fixes # .

- 
- 
-


bonustrack:
Can we make the config var optional so we don't enable it on testnet hub?

wa0x6e:
Done, envelopDB connection is now optional

ChaituVR:
> Done, envelopDB connection is now optional

I think we need to address https://github.com/snapshot-labs/snapshot-sequencer/pull/470 first no?

PR #952, opened 2024-11-04
fix: fix: update schema to reflect new columns in proposals

wa0x6e:
Update schema.sql file to reflect new columns in proposals


src/helpers/schema.sql
```diff
@@ -58,6 +58,9 @@ CREATE TABLE proposals (
   scores_state VARCHAR(24) NOT NULL,
   scores_total DECIMAL(64,30) NOT NULL,
   scores_updated INT(11) NOT NULL,
+  vp_usd decimal(64,30) NOT NULL DEFAULT '0.000000000000000000000000000000',
+  vp_eth decimal(64,30) NOT NULL DEFAULT '0.000000000000000000000000000000',
+  vp_value_by_strategy json NOT NULL,
```
ChaituVR, line 63: These cols are only used internally, not used in our backend/testnet. so not sure if we want everyone to use these to hold default values

@bonustrack wyt?

ChaituVR (changes requested):

bonustrack:
Please don't merge, I'll do some change in the columns, removing vp_usd and vp_eth and add scores_total_value also add a column in votes. These columns will later be exposed in the API.

src/helpers/schema.sql
bonustrack: Whats wrong about default value?

wa0x6e:
See https://github.com/snapshot-labs/snapshot-hub/pull/970

PR #964, opened 2024-12-19
fix: wrong counts

ChaituVR:
Fixes https://github.com/snapshot-labs/workflow/issues/341

### Summary:
Inside `loadSpacesMetrics`, we were using previous counts if metrics doesn't contain certain properties
```ts
spacesMetadata[space].counts = {
    ...spacesMetadata[space].counts,
    ...metrics
};
```  
- If there are no active proposals then `getProposals` return an object without `activeProposals` property
- If there are no active proposals / recent proposals, then `getProposals` returns empty object
- In these cases, it will show previous counts
- This PR fixes that and uses new counts

### How to test:
- Create a proposal that ends in 5 mins
- Try the following query
```graphql
{
  space(id: "thanku.eth") {
    id
    activeProposals
    proposalsCount1d
    proposalsCount7d
    proposalsCount30d
  }
}
```
- `activeProposals` should be `1`
- Once the proposal end `activeProposals` should be `0`

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -263,6 +263,7 @@ async function loadSpacesMetrics() {
     getProposals(),
     getVotes()
   ]);
+  
   const [followerMetrics, proposalMetrics, voteMetrics] = results;
   Object.keys(spacesMetadata).forEach(space => {
     spacesMetadata[space].counts = {
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -167,7 +167,15 @@ async function loadSpaces() {
 }
 
 async function getProposals(): Promise<
-  Record<string, { activeProposals: number; proposalsCount7d: number }>
+  Record<
+    string,
+    {
+      activeProposals: number;
+      proposalsCount1d: number;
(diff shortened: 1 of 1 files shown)


src/helpers/spaces.ts
```diff
@@ -250,19 +258,22 @@ async function getFollowers(): Promise<
...
-        ...metrics
-      };
-    }
-    log.info(`[spaces] ${metricFn.name.replace('get', '')} metrics loaded`);
+  const results = await Promise.all([
+    getFollowers(),
+    getProposals(),
+    getVotes()
+  ]);
+  const [followerMetrics, proposalMetrics, voteMetrics] = results;
```
bonustrack, line 266: We should add new line if there is `await` call
```suggestion
  ]);
  
  const [followerMetrics, proposalMetrics, voteMetrics] = results;
```

bonustrack (approved):
utACK

src/helpers/spaces.ts
ChaituVR: we have this in many places, better to add a eslint rule

PR #975, opened 2025-01-23
fix: include networks from delegation strategies

ChaituVR:
Issue:
On the v1 interface it says that there are 10 spaces on the Sonic network but on the v2 interface i only see 7 spaces.

The other three spaces

`heyanondao.eth`: enabled "Hide space from homepage" in settings
`beefydao.eth` and `profit.beefy.eth`: These spaces use this network inside delegation strategy, this PR adds these spaces to response

### How to test:
Use following query, it should return 7 spaces before and 9 spaces after fix
```Graphql
{
  ranking(where: {network: "146"}) {
    metrics {
      total
      categories
    }
    items {
      id
      network
      strategies {
        name
        network
        params
      }
    }
  }
}

```

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -91,11 +91,15 @@ function sortSpaces() {
 
 function mapSpaces() {
   Object.entries(spaces).forEach(([id, space]: any) => {
-    const networks = uniq(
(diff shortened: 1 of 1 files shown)


wa0x6e (approved):
tAck

PR #976, opened 2025-01-27
feat: return premium property for networks

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/389
See https://github.com/snapshot-labs/sx-monorepo/pull/1120#discussion_r1930677289

This PR returns the `premium` property for each network.

For this, a new table will be needed to hold network metadata:

```sql
CREATE TABLE networks (
  id VARCHAR(64) NOT NULL,
  premium SMALLINT UNSIGNED NOT NULL DEFAULT '0',
  PRIMARY KEY (id),
  INDEX premium (premium)
);
```

For now, the table will only contain data for premium network (missing network in this table means non-premium network)

See https://github.com/snapshot-labs/workflow/issues/389#issuecomment-2616078329 for the SQL to import data

### Test

Test with this graphql query 

```graphql
query { 
networks { 
id
spacesCount
premium}
}
```

Which should return something like

```json
{
  "data": {
    "networks": [
      {
        "id": "1",
        "spacesCount": 79904,
        "premium": false
      },
      {
        "id": "3",
        "spacesCount": 36,
        "premium": false
      },
...
```

### Note 

Later, we could move more data to this table, like the `spaces_count`, and all the networks in this table, not just the premiun ones

### DIFF
--- a/src/graphql/operations/networks.ts
+++ b/src/graphql/operations/networks.ts
@@ -1,6 +1,7 @@
+import db from '../../helpers/mysql';
 import { spaces } from '../../helpers/spaces';
 
-export default function () {
+export default async function () {
   const networks: any = Object.values(spaces).reduce((acc: any, space: any) => {
     const spaceNetworks = [
       space.network,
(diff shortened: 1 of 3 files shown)


src/graphql/operations/networks.ts
```diff
@@ -22,8 +23,23 @@ export default function () {
...
+    return acc;
+  }, {});
+
+  const premiumNetworks = await db.queryAsync(
+    'SELECT * FROM networks WHERE premium = 1'
+  );
+  premiumNetworks.forEach((network: any) => {
+    results[network.id] ||= { id: network.id, spacesCount: 0 };
+    results[network.id].premium = true;
+  });
```
ChaituVR, line 42: This will display a network without spaces on Explore page, we wanted to display only networks with spaces right?

ChaituVR (approved):
tAck

src/graphql/operations/networks.ts
ChaituVR: Usually new premium networks will have `0` spacesCount

ChaituVR: hmm sorry i missed it

PR #981, opened 2025-02-13
fix: reduce popularty of turbo subspaces

ChaituVR:
Increases popularity of turbo space when space doesn't have a parent space


bonustrack (approved):
utACK

PR #986, opened 2025-03-03
feat: add network names

bonustrack:
- I've added all the networks (based on spaces usage) to the livenet and testnet db. Now when we are adding a new network, it should also be added in the dbs. 
- The GraphQL API now expose the network name. 
- Also simplified the logic in the `networks` GraphQL operation, we are not the large reduce function on each request anymore but only once.


ChaituVR:
> * I've added all the networks (based on spaces usage) to the livenet and testnet db. 

- I see many empty names on livenet 🤔 
image


- Why only based on space activity? I think we should add all networks and filter if no space is using it 🤔 Because if someone uses a network now with v1 or snapshot.js it will miss in this list, can I add all missing networks?

src/helpers/schema.sql
```diff
@@ -216,6 +216,7 @@ CREATE TABLE skins (
 
 CREATE TABLE networks (
   id VARCHAR(64) NOT NULL,
+  name VARCHAR(32) NOT NULL,
```
ChaituVR, line 219: Looks like we also have `testnet` column now.

bonustrack:
> > * I've added all the networks (based on spaces usage) to the livenet and testnet db.
> 
> * I see many empty names on livenet 🤔
> 
> image
> * Why only based on space activity? I think we should add all networks and filter if no space is using it 🤔 Because if someone uses a network now with v1 or snapshot.js it will miss in this list, can I add all missing networks?

Takes quite some time to fill network names, I've started, will complete overtime. Names aren't used atm

bonustrack:
@ChaituVR I've added all the missing network in production (not sure how we go about this on testnet), and also added name for all the networks

ChaituVR (changes requested):
- Any reasons for using different names than networks.json? 
For example `Harmony Shard 0` instead of `Harmony `, `PHI Network v2` on DB instead of `PHI Network` etc

I see some differences with networks.json and networks in DB
- `33111` is missing in DB but exists on networks.json
- Also I see there is these networks with id `900000`, `888888` and `80` on DB but not in networks.json

ChaituVR:
There are 65 differences with networks.json and DB networks
```ts
[
    81,
    97,
    157,
    300,
    1001,
    1072,
    1328,
    1663,
    2221,
    2522,
    5611,
    6102,
    9990,
    10243,
    13337,
    13473,
    14800,
    33111,
    43113,
    57054,
    59141,
    80001,
    80002,
    84531,
    84532,
    686868,
    713715,
    763373,
    810181,
    11155111,
    11155420,
    111557560,
    168587773,
    245022926,
    1666700000,
    106,
    11297108109,
    119,
    128,
    1319,
    1701,
    188,
    20,
    2020,
    278611351,
    288,
    32659,
    333999,
    361,
    42,
    42262,
    43288,
    44,
    499,
    50,
    534,
    7,
    70,
    70001,
    7700,
    80,
    87,
    888888,
    900000,
    99
]
```

I think these are deleted networks which exist before

bonustrack:
@ChaituVR Are you talking about mainnet or testnet db? On mainnet db I've added all the networks that are currently used by a space, and all the networks from networks.json, removing all the testnet networks.

ChaituVR:
> @ChaituVR Are you talking about mainnet or testnet db? On mainnet db I've added all the networks that are currently used by a space, and all the networks from networks.json, removing all the testnet networks.

I am talking about mainnet DB, looks like you've added all networks based on space usage but many spaces still use deleted networks. instead, we should use ones from networks.json

bonustrack:
@ChaituVR You can takeover this and keep only the networks from networks.json

bonustrack:
@ChaituVR I did the requested changes already

PR #989, opened 2025-03-25
feat: turbo_expiration_timestamp

pscott:
See: https://github.com/snapshot-labs/workflow/issues/488#issue-2902688159

- Adds a `turbo_expiration_timestamp` in the schema.sql
- Adds a `turbo_expiration_timestamp` in the schema.gql

### DIFF
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -437,6 +437,7 @@ type Space {
   flagged: Boolean
   hibernated: Boolean
   turbo: Boolean
+  turbo_expiration_timestamp: Int
   rank: Float
   boost: BoostSettings
   created: Int!
--- a/src/graphql/schema.gql
+++ b/src/graphql/schema.gql
@@ -437,7 +437,7 @@ type Space {
   flagged: Boolean
   hibernated: Boolean
   turbo: Boolean
-  turbo_expiration_timestamp: Int
+  turbo_expiration: Int
   rank: Float
   boost: BoostSettings
   created: Int!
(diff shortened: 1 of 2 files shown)


src/graphql/schema.gql
```diff
@@ -437,6 +437,7 @@ type Space {
   flagged: Boolean
   hibernated: Boolean
   turbo: Boolean
+  turbo_expiration_timestamp: Int
```
ChaituVR, line 440: maybe just
```suggestion
  turbo_expiration: Int
```

src/helpers/schema.sql
```diff
@@ -7,6 +7,7 @@ CREATE TABLE spaces (
   flagged INT NOT NULL DEFAULT '0',
   hibernated INT NOT NULL DEFAULT '0',
   turbo INT NOT NULL DEFAULT '0',
+  turbo_expiration_timestamp INT NOT NULL DEFAULT '0',
```
ChaituVR, line 10:
```suggestion
  turbo_expiration INT NOT NULL DEFAULT '0',
```

ChaituVR (approved):
utAck

src/helpers/schema.sql
```diff
@@ -7,6 +7,7 @@ CREATE TABLE spaces (
   flagged INT NOT NULL DEFAULT '0',
   hibernated INT NOT NULL DEFAULT '0',
   turbo INT NOT NULL DEFAULT '0',
+  turbo_expiration INT NOT NULL DEFAULT '0',
```
ChaituVR, line 10: Sorry didn't notice it earlier, maybe we can use the same type as `created` and `updated`?

```sql
  created BIGINT NOT NULL,
  updated BIGINT NOT NULL,
 ```

pscott: yeah, i didn't know int / bigint were not signed in sql. updated to use bigint so we can have 4 bytes of positive range : f655ce85ea0aa9cc64dd312ab7f0d2b0c4f2dc9e

PR #999, opened 2025-04-17
feat: use turbo expiration field to deduce turbo status

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/420

Use the new `turbo_expiration` field to deduce a space's turbo status.

In case expiration is empty, will fallback to the previous `turbo` field for backward compatibility.

This is a soft migration, once all the existing turbo spaces have a proper expiration date, we will remove the `turbo` field for good


ChaituVR (approved):
utAck

PR #1001, opened 2025-04-23
refactor: avoid recursive calls

wa0x6e:
Toward https://github.com/snapshot-labs/workflow/issues/544

Migrate away from recursive calls


ChaituVR (approved):
utAck

Issue #1005, opened 2025-04-28
bug: homepage not returning correct turbo state


wa0x6e:
see https://testnet.snapshot.box/#/explore

only 3 spaces are shown as turbo, whereas more spaces than that have turbo

PR #1006, opened 2025-04-28
fix: infer turbo status from turbo expiration date

wa0x6e:
Fixes https://github.com/snapshot-labs/snapshot-hub/issues/1005

This PR fix an issue where a space is not considered as turbo, even if the turbo expiration date is in the future

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -57,7 +57,7 @@ type Metadata = {
 };
 
 function isTurbo(turbo: number, turboExpiration: number): boolean {
-  return turbo === 1 || turboExpiration > Date.now();
+  return turbo === 1 || turboExpiration > Date.now() / 1e3;
 }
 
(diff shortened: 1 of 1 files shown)


src/helpers/spaces.ts
```diff
@@ -56,6 +56,10 @@ type Metadata = {
   pluginNames: string[];
 };
 
+function isTurbo(turbo: number, turboExpiration: number): boolean {
+  return turbo === 1 || turboExpiration > Date.now() / 1e3;
```
ChaituVR, line 60: Maybe we can just use `turboExpiration > Date.now() / 1e3` ? i remember at some point, the plan was to deprecate `turbo` right?

wa0x6e: Yes, but until we do, we still need this for the transition period, as we're filling up the schnaps db with data

PR #1008, opened 2025-05-01
fix: remove usage of global `spaces` object

wa0x6e:
Closes https://github.com/snapshot-labs/workflow/issues/544

Remove usage and delete the global `spaces` object from helpers/spaces.ts

This object is huge, and contains all the spaces settings, and is also 90% redundant with the `spacesMetadata` object.

Run 

```graphql
query { 
strategies {
  id
  spacesCount
}
}
```

It should return the same results as before this PR. Same with `skins`, `plugins` and `validations` entity.

- before PR app memory usage: 265mb
- after PR app memory usage: 128mb

### DIFF
--- a/src/helpers/spaces.ts
+++ b/src/helpers/spaces.ts
@@ -19,8 +19,6 @@ const TESTNET_NETWORKS = (
   .filter(network => network.testnet)
   .map(network => network.key);
 
-export let spaces = {};
-
 export let rankedSpaces: Metadata[] = [];
 
 export let networkSpaceCounts: Record<string, number> = {};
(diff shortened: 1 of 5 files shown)


src/helpers/spaces.ts
```diff
@@ -100,7 +100,7 @@ function sortSpaces() {
     });
 }
 
-function mapSpaces() {
+function mapSpaces(spaces: Record<string, Space>) {
```
ChaituVR, line 103: `Cannot find name 'Space'.`

wa0x6e, line 103:
```suggestion
function mapSpaces(spaces: Record<string, any>) {
```

wa0x6e:
> Looks good to me

Let’s try this one first, if it resolves, will not need to move back to docker

ChaituVR (approved):
tAck

PR #1009, opened 2025-05-01
fix: docker build

ChaituVR:
related to https://github.com/snapshot-labs/workflow/issues/544

### How to test:
Run `docker build -t snapshot-hub .` which was failing before with error:

```
Package 'python' has no installation candidate
```

### TODO:
- [x] On DO, we need to add back this line and remove extra `-` before env
Untitled 4


wa0x6e:
I think we may also need to add back the github actions

ChaituVR:
> I think we may also need to add back the github actions

Do we need to publish to docker 🤔 not sure where we use it

wa0x6e (approved):
utAck

PR #1011, opened 2025-05-08
fix: format spaces with turbo expiration

ChaituVR:
### How to test:

```GraphQL
{
  ranking(where: {search: "chaingov.eth"}) {
    items {
      id
      turbo
      turboExpiration
    }
  }
  space(id: "chaingov.eth") {
    id
    turbo
    turboExpiration
  }
}
```

Before it should return false for turbo, after it should return true


wa0x6e (approved):
utAck

PR #1019, opened 2025-06-20
perf: optimize messages query

wa0x6e:
Fixes https://github.com/snapshot-labs/workflow/issues/585

The messages fetching query is using the `timestamp` index, instead of the `space` index, which is more appropriate when filtering by spaces.

This PR forces a query with spaces filtering to use the space index.

### Test

The same graphql mentioned in the issue should resolve in less than 5s

### Note

It's not using multiple index in the `USE INDEX()` (even it supported), from my tests, only using 2 index is as slow as before this PR


ChaituVR (approved):
tAck

PR #1024, opened 2025-07-09
perf: improve messages queries

wa0x6e:
Fixes https://github.com/snapshot-labs/workflow/issues/585

This PR improves upon the previous fix.

We will now use the query hint `use index` when the query is filtered by spaces, and not ordered by `mci` (which is an indexed column).

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "@snapshot-labs/keycard": "^0.5.1",
     "@snapshot-labs/snapshot-metrics": "^1.4.1",
     "@snapshot-labs/snapshot-sentry": "^1.5.5",
-    "@snapshot-labs/snapshot.js": "^0.14.2",
+    "@snapshot-labs/snapshot.js": "^0.14.7",
     "bluebird": "^3.7.2",
     "connection-string": "^1.0.1",
     "cors": "^2.8.5",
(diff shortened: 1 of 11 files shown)


ChaituVR:
I tried query with 
```graphql
query BatchLatestMCI {
_a42edf54dd93220a6f8a410c0919c77a67d2feef8a8295123eabe17426853284: messages(where: {
      space: "swarmfoundation.eth",
      type_in: ["proposal", "delete-proposal"],
      },
      first: 1,
      orderBy: "mci",
      orderDirection: desc) {
        mci
      }
_f88a5f7ff244b35c39427d542ba718362b413eec175276b193d459404dd1d0fc: messages(where: {
      space: "etherfi-dao.eth",
      type_in: ["proposal", "delete-proposal"],
      },
      first: 1,
      orderBy: "mci",
      orderDirection: desc) {
        mci
      }
}
```

it is still slow somehow 🤔 taking forever, is it same for you?

wa0x6e:
Aside from that, nothing really can be done, beside adding more index

wa0x6e:
Mci index should already be the one used by default, since it's the primary key. Do you notice difference by using it explicitly?

ChaituVR:
hmm looks like above queries are resolving fast on hub.snapshot.org 😅 i tried different spaces and they are working as well. weird

PR #1029, opened 2025-07-25
fix: remove undefined keys from the global networks count variable

wa0x6e:
This PR fix an issue where the global `networkSpacesCount` variable contains the key `undefined` as network id, and was counting 30k spaces in it.

This PR fix the issue by ignoring those undefined values

Also improves the metrics, by using the correct global variable to retrieve those stats instead 
 looping the spaces object
### Test

- For the lines starting by `spaces_per_network_count`, no more `undefined` network


ChaituVR (approved):
utAck

PR #1030, opened 2025-07-25
fix: use new expiration column to return turbo status

wa0x6e:
This metrics is still using the old `turbo` column to infer a space turbo status.

Since the introduction of the auto-payment, status is inferred with the `turbo_expiration` column, the old column is not updated anymore.

This PR uses the new col to get the turbo status, and is replacing the old `turbo` metrics by 2 new ones:

- `active_turbo`
- `expired_turbo`

### How to test

- You should see something like this:

```
# HELP spaces_per_status_count Number of spaces per status
# TYPE spaces_per_status_count gauge
spaces_per_status_count{status="verified"} 775
spaces_per_status_count{status="flagged"} 11508
spaces_per_status_count{status="hibernated"} 6341
spaces_per_status_count{status="active_turbo"} 28
spaces_per_status_count{status="expired_turbo"} 3
```

### DIFF
--- a/src/helpers/metrics.ts
+++ b/src/helpers/metrics.ts
@@ -95,15 +95,24 @@ new client.Gauge({
   help: 'Number of spaces per status',
   labelNames: ['status'],
   async collect() {
-    ['verified', 'flagged', 'turbo', 'hibernated'].forEach(async status => {
-      this.set(
-        { status },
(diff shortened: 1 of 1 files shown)


ChaituVR (approved):
utAck

PR #1036, opened 2025-07-29
fix: use fetch from snapshot.js

wa0x6e:
From https://github.com/snapshot-labs/snapshot-hub/pull/1031#discussion_r2239949214

Use SDK from all fetch related calls, which support timeout

The timeout use will hopefully fixes any potential hanging request, making the strategies refresh loop hanging

### Test

```
query Test {
spaces(where:{ controller: "0x370dE8E14fd6F054b17220a3882beF2960a159F3" } ) {
  id 
  children {
    id 
  }
  parent {
    id
  }
}
}
```

It should work


ChaituVR (approved):
utAck

PR #1039, opened 2025-08-10
chore: enable SSL with rejectUnauthorized

bonustrack:
**DO NOT MERGE**

### DIFF
--- a/src/helpers/mysql.ts
+++ b/src/helpers/mysql.ts
@@ -19,6 +19,7 @@ hubConfig.connectTimeout = 60e3;
 hubConfig.acquireTimeout = 60e3;
 hubConfig.timeout = 60e3;
 hubConfig.charset = 'utf8mb4';
+hubConfig.ssl = { rejectUnauthorized: true };
 
 const hubDB = mysql.createPool(hubConfig);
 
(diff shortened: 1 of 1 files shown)


src/helpers/mysql.ts
```diff
@@ -19,6 +19,7 @@ hubConfig.connectTimeout = 60e3;
 hubConfig.acquireTimeout = 60e3;
 hubConfig.timeout = 60e3;
 hubConfig.charset = 'utf8mb4';
+hubConfig.ssl = { rejectUnauthorized: true };
```
ChaituVR, line 22: For tests to pass

```suggestion
if (process.env.NODE_ENV === 'production') {
  hubConfig.ssl = { rejectUnauthorized: true };
}
```

src/helpers/mysql.ts
```diff
@@ -33,6 +34,8 @@ sequencerConfig.connectTimeout = 60e3;
 sequencerConfig.acquireTimeout = 60e3;
 sequencerConfig.timeout = 60e3;
 sequencerConfig.charset = 'utf8mb4';
+sequencerConfig.ssl = { rejectUnauthorized: true };
```
ChaituVR, line 37:
```suggestion
```suggestion
if (process.env.NODE_ENV === 'production') {
  sequencerConfig.ssl = { rejectUnauthorized: true };
}
```
```

PR #1048, opened 2025-08-17
feat: default limit for all _in fields

ChaituVR:
Right now we don't have any limits for _in fields, so any one can pass many items to these

### Summary
- added default limit for all *_in fields to ARG_LIMITS

### How to test
- Previous limits should still work
- Update proposals query limit here https://github.com/snapshot-labs/snapshot-hub/blob/feat-default-in-field-limit/src/graphql/helpers.ts#L28 
- Query with
```Graphql
{
  proposals(where:{space_in:["ens.eth",............]}){
    id
  }
}
```

### DIFF
--- a/src/graphql/helpers.ts
+++ b/src/graphql/helpers.ts
@@ -21,7 +21,8 @@ export class PublicError extends Error {}
 const ARG_LIMITS = {
   default: {
     first: 1000,
-    skip: 5000
+    skip: 5000,
+    '*_in': 1000 // Default limit for all _in arguments
   },
   proposals: {
     space_in: 10000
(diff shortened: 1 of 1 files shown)


src/graphql/helpers.ts
```diff
@@ -21,7 +21,8 @@ export class PublicError extends Error {}
 const ARG_LIMITS = {
   default: {
     first: 1000,
-    skip: 5000
+    skip: 5000,
+    '*_in': 1000 // Default limit for all _in arguments
```
bonustrack, line 25: This is too high, dont think currently anyone hit this limit, I'd just use 100 here. 
```suggestion
    '*_in': 100 // Default limit for all _in arguments
```

src/graphql/helpers.ts
```diff
@@ -53,6 +54,13 @@ export function checkLimits(args: any = {}, type) {
   const { where = {} } = args;
   const typeLimits = { ...ARG_LIMITS.default, ...(ARG_LIMITS[type] || {}) };
 
+  // overwrite default limit for all *_in fields
```
bonustrack, line 57: We should also limit the `_not_in` fields

bonustrack:
What happen if we hit the limit? It will continue with the max? I think it should throw an error

src/graphql/helpers.ts
ChaituVR: Looks like it already limits `_not_in` 😅 

```suggestion
  // overwrite default limit for all *_in or *_not_in fields
```

ChaituVR:
> What happen if we hit the limit? It will continue with the max? I think it should throw an error

It will throw an error, example: "The `app_not_in` argument must not be greater than 100"

bonustrack (approved):
utACK

PR #1075, opened 2026-05-04
feat: migrate from Yarn to Bun

bonustrack:
## Summary

Migrate the package manager from Yarn to Bun. Final shape mirrors **sx-monorepo's `apps/api`**: Bun for install + build, Node for runtime (dev, prod, tests). This is the target shape because the medium-term plan is to move snapshot-hub into sx-monorepo, so it makes sense to align now.

## Concrete changes

- **`package.json`**: `packageManager: "bun@1.2.0"`, `yarn` calls replaced by `bun run`. `dev` uses `nodemon`, `start` uses `node` (unchanged from master), `test:setup` uses `ts-node`. `engines.node` lowered to `>=22.6.0` (more on that below).
- **Lockfile**: `yarn.lock` removed, `bun.lock` committed. The new lockfile resolves newer versions of eslint/prettier plugins which surfaced a few unused-catch-binding lint errors and one prettier autofix — fixed in-PR.
- **Workflows**: `lint`, `test`, `create-sentry-release` now pass `package_manager: bun` to the shared reusable workflows. Depends on snapshot-labs/actions#22.
- **No Docker**: `Dockerfile`, `docker-compose.yml`, `.dockerignore`, `publish-docker-image.yml` removed. Production deploys via DO's Bun buildpack.
- **No Bun runtime code**: no `bunShim`, no mysql `_startTLS` monkey-patch. Node runtime everywhere makes those unnecessary.

## Choices made along the way

A few decisions are worth recording because they cost time during the migration:

1. **Bun runtime → Node runtime.** First attempt ran the production server on Bun. That hit two compat issues: `mysql`'s TLS upgrade path uses Node-internal APIs Bun doesn't implement, and `prom-client` calls `perf_hooks.monitorEventLoopDelay` which throws on Bun. We had monkey-patches for both. They're gone now — the cleaner path is to keep Node as the runtime everywhere (matches sx-monorepo).

2. **DO Bun buildpack version drift.** When we ran on Bun, the DO buildpack v0.0.2 ignored our pinned `bun@1.2.0` and shipped its own. Switching the runtime to Node sidesteps this — Bun's only role is install/build, so its version no longer affects runtime behavior.

3. **`engines.node: ">=22.6.0"` (was `>=24.14` on master).** DO's Bun buildpack manifest doesn't carry Node 24 — `>=24.14.0` returns "No result". Node 22.6 is the closest LTS the buildpack accepts and is what sx-monorepo pins. Master can keep Node 24 because it deploys via the larger Node buildpack (yarn.lock).

4. **Kept `mysql` (didn't switch to `mysql2`).** Briefly tried mysql2 to avoid the TLS monkey-patch, then reverted: dropping Bun runtime makes the patch unnecessary and lets us keep the legacy `mysql` package without changes to ~30 callsites.

## Prerequisites before merge

- [ ] **snapshot-labs/actions#22** must merge first (adds the `package_manager` input).
- [ ] Flip the three workflow `uses:` refs from `@feat/bun-support` back to `@main` after the above merges.
- [ ] DO production app spec must drop `dockerfile_path: Dockerfile` (already done by maintainer in DO dashboard).

## Test plan

- [x] `bun install` reproduces `bun.lock`.
- [x] `bun run lint` / `bun run typecheck` / `bun run build` pass.
- [x] CI Lint + Test workflows green on this branch.
- [x] DO production deploy succeeds and serves traffic on Node 22 + Bun-buildpack-built artifacts.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/graphql/helpers.ts
```diff
@@ -216,7 +218,7 @@ export function buildWhereQuery(
           where[key] = Array.isArray(where[key])
             ? formattedAddresses
             : formattedAddresses[0];
-        } catch (e: any) {
+        } catch {
```
wa0x6e, line 217: Those changes are unrelated to bun migration

wa0x6e (changes requested):

test/setup.ts
```diff
@@ -24,7 +24,7 @@ const setup = async () => {
     }
   } catch (e: any) {
     if (e.code === 'ER_BAD_DB_ERROR') {
-      console.error('Test database not setup, please run `yarn test:setup`');
+      console.error('Test database not setup, please run `bun test:setup`');
```
wa0x6e: It's `bun run test:setup`

package.json
```diff
@@ -4,25 +4,23 @@
...
-    "test:e2e": "PORT=3030 start-server-and-test 'yarn start:test' http://localhost:3030 'dotenv -e test/.env.test jest --runInBand --collectCoverage=false test/e2e/'"
+    "test:e2e": "PORT=3030 start-server-and-test 'bun run start:test' http://localhost:3030 'dotenv -e test/.env.test jest --runInBand --collectCoverage=false test/e2e/'"
   },
   "eslintConfig": {
     "extends": "@snapshot-labs"
   },
   "prettier": "@snapshot-labs/prettier-config",
-  "engines": {
-    "node": ">=24.14"
-  },
```
wa0x6e, line 25: Why was this removed ?

test/setup.ts
bonustrack: Fixed in 199b13f — same footgun as in the reusable workflows (`bun <name>` is fine for arbitrary script names but conflicts with built-ins like `test`, so `bun run` is safer everywhere).

src/graphql/helpers.ts
bonustrack: Fair callout — these aren't part of the migration concept, but they were forced by it: the new `bun.lock` resolved `eslint@8.57.1` and `@typescript-eslint/*` to newer versions than the old `yarn.lock` had pinned, and those flag unused catch bindings as errors. Without these fixes the Lint workflow fails on this branch.

Happy to split them into a follow-up PR before merging this one if you'd prefer to keep the diff scoped strictly to the package-manager swap. Let me know.

package.json
bonustrack: Because we now run on Bun (the runtime, not just the package manager) and there's no Node in the deployed image. Keeping `engines.node` made DigitalOcean's Bun buildpack try to install a Node version from its (limited) manifest and fail with `Could not parse Range` / `No result`.

If we want to gate Bun versions instead, that's what `packageManager: bun@1.2.0` does. Or we could add `engines.bun` if you'd like an explicit floor — let me know.

Dockerfile
```diff
@@ -5,7 +5,12 @@ FROM node:24.14-slim
 RUN apt-get update && apt-get upgrade -y
 
 # Install required O.S. packages
-RUN apt-get install -y git python3 make g++
+RUN apt-get install -y git python3 make g++ curl unzip
+
+# Install Bun
+RUN curl -fsSL https://bun.sh/install | bash -s -- --no-modify-path \
```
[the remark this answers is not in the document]
bonustrack: This is now moot — the Dockerfile no longer installs Bun via curl. As of 8dbff7f the base is `oven/bun:1.2.0-slim`, which pins both the runtime and the Bun version to a tag (and matches `packageManager: bun@1.2.0`).

.github/workflows/lint.yml
```diff
@@ -7,7 +7,8 @@ jobs:
     strategy:
       matrix:
         target: ['24.14']
-    uses: snapshot-labs/actions/.github/workflows/lint.yml@main
+    uses: snapshot-labs/actions/.github/workflows/lint.yml@feat/bun-support
```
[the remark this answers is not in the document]
bonustrack: Correct — this is intentionally temporary. snapshot-labs/actions#22 (the PR adding the `package_manager: bun` input) hasn't merged yet, so I pointed the three workflow refs (`lint`, `test`, `create-sentry-release`) at the feature branch so CI on this PR could go green end-to-end.

**Before merging this PR**, all three need to flip back to `@main`. Order: merge actions#22 → flip refs here → merge this PR.

.github/workflows/test.yml
```diff
@@ -7,8 +7,9 @@ jobs:
     strategy:
       matrix:
         target: ['24.14']
-    uses: snapshot-labs/actions/.github/workflows/test.yml@main
+    uses: snapshot-labs/actions/.github/workflows/test.yml@feat/bun-support
```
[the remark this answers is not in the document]
bonustrack: Correct — this is intentionally temporary. snapshot-labs/actions#22 (the PR adding the `package_manager: bun` input) hasn't merged yet, so I pointed the three workflow refs (`lint`, `test`, `create-sentry-release`) at the feature branch so CI on this PR could go green end-to-end.

**Before merging this PR**, all three need to flip back to `@main`. Order: merge actions#22 → flip refs here → merge this PR.

.github/workflows/create-sentry-release.yml
```diff
@@ -8,8 +8,9 @@ jobs:
     strategy:
       matrix:
         target: ['24.14']
-    uses: snapshot-labs/actions/.github/workflows/create-sentry-release.yml@main
+    uses: snapshot-labs/actions/.github/workflows/create-sentry-release.yml@feat/bun-support
```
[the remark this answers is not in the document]
bonustrack: Correct — this is intentionally temporary. snapshot-labs/actions#22 (the PR adding the `package_manager: bun` input) hasn't merged yet, so I pointed the three workflow refs (`lint`, `test`, `create-sentry-release`) at the feature branch so CI on this PR could go green end-to-end.

**Before merging this PR**, all three need to flip back to `@main`. Order: merge actions#22 → flip refs here → merge this PR.

bonustrack:
This PR is deployed in production in livenet and testnet already

wa0x6e:
We may want to merge https://github.com/snapshot-labs/snapshot-hub/pull/1076 first, to avoid all the unrelated lint fix in this PR

package.json
wa0x6e: Since we're now back to `node` from https://github.com/snapshot-labs/snapshot-hub/pull/1075/commits/68da441ac31d8a0502f76fe6b5df57adcb86a0e0 , the `engines` should be relevant again

package.json
```diff
@@ -4,25 +4,23 @@
...
     "typecheck": "tsc --noEmit",
-    "dev": "nodemon src/index.ts",
+    "dev": "bun --hot src/index.ts",
     "build": "tsc && copyfiles src/graphql/*.gql dist/",
     "start": "node dist/src/index.js",
-    "start:test": "dotenv -e test/.env.test yarn dev",
-    "test:setup": "dotenv -e test/.env.test yarn ts-node ./test/setupDb.ts",
-    "test": "yarn test:setup; PORT=3030 start-server-and-test 'yarn start:test' http://localhost:3030 'dotenv -e test/.env.test jest --runInBand'",
+    "start:test": "dotenv -e test/.env.test bun run dev",
+    "test:setup": "dotenv -e test/.env.test bun ./test/setupDb.ts",
```
wa0x6e: Does this work with just `bun`, or does this need `bunx` like on sequencer ? https://github.com/snapshot-labs/snapshot-sequencer/pull/638/changes#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R13

bonustrack:
@copilot resolve the merge conflicts in this pull request

wa0x6e:
@copilot resolve the merge conflict

wa0x6e (approved):
tAck

